PR #1, opened 2018-01-30
added config and logging

nitzantomer:


scripts/src/logging.ts
```diff
@@ -0,0 +1,72 @@
+import * as winston from "winston";
+import {GenericTextTransportOptions} from "winston";
+import {GenericTransportOptions} from "winston";
+import {FileTransportOptions} from "winston";
+
+export interface LogTarget {
+	name: string;
+	type: "console" | "file";
+	level: "debug" | "info" | "warn" | "error";
+	timestamp?: boolean | (() => string | boolean);
```
doodyparizada: I don't understand this. is timestamp a boolean?

scripts/src/logging.ts
```diff
@@ -0,0 +1,72 @@
...
+export function getLogger(...targets: LogTarget[]): winston.LoggerInstance {
+	const options: winston.LoggerOptions = {};
+	options.transports = targets.map(target => createTarget(target));
+
+	return new winston.Logger(options);
+}
+
+type WinstonTransportOptions = GenericTransportOptions & GenericTextTransportOptions;
+function createTarget(target: LogTarget): winston.TransportInstance {
+	let cls: { new (options: WinstonTransportOptions): winston.TransportInstance };
```
doodyparizada: I don't know what this means

scripts/src/logging.ts
```diff
@@ -0,0 +1,72 @@
...
+			break;
+
+		default:
+			throw new Error("unsupported log target type: " + target.type);
+	}
+
+	return new cls(mergeOptions(defaults, options));
+}
+
+function mergeOptions(defaults: WinstonTransportOptions, options: WinstonTransportOptions): WinstonTransportOptions {
```
doodyparizada: Is there a better way?

scripts/src/logging.ts
nitzantomer: `timestamp` can be either a boolean or a function which returns either a boolean or a string.

scripts/src/logging.ts
nitzantomer: `cls` is a class, that is, a constructor function which expects one param (`options`) and returns a new instance of `winston.TransportInstance`.
notice that later there's:
`new cls(options)`

scripts/src/logging.ts
nitzantomer: a better way?
what's wrong with this one?

scripts/src/logging.ts
doodyparizada: why is it a boolean? why not string or time in seconds from epoch?

nitzantomer: it is only to set whether or not a timestamp should be included.
if so, the timestamp is added.  it's not the format.

scripts/src/logging.ts
doodyparizada: a more generic, merge 2 objects

PR #2, opened 2018-01-31
add a draft of DB models using sequelize

doodyparizada:
I'm missing the order class which will be implemented as in a redis cache as discussed and will lock a offer instance to a user for a period of time.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,11 @@
         "@types/node": "9.4.0"
       }
     },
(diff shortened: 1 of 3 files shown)


nitzantomer:
Looks good.
But, we want to keep some conventions.
So far the code used double quotes (`"`) and not singles (`'`), also it uses tabs and not spaces for indentation.
We can talk about the conventions and change later, but for now, let's keep it all the same

PR #3, opened 2018-02-04
add mock offers

doodyparizada:


scripts/src/services/offers.ts
```diff
@@ -53,10 +53,45 @@ export type OfferList = {
 }
 
 export const getOffers = async (options): Promise<ServiceResult<OfferList>> => {
+
+	const earn = [
```
nitzantomer: consts can move to be outside of the function

doodyparizada: The earn and spend vars are temporary. What about the IMAGE_BASE, should that be in a separate const file that depends on stage/prod environment?

nitzantomer: I created the config directory for that.
Currently, we only have the [default config](https://github.com/kinfoundation/marketplace-server/blob/master/config/default.json) but we can create different once there.

And the image base const can be added.  If you wanna do that then add it here:
https://github.com/kinfoundation/marketplace-server/blob/master/scripts/src/config.ts#L3

PR #4, opened 2018-02-07
syncing the models

nitzantomer:


scripts/src/models/all.ts
```diff
@@ -1,83 +1,111 @@
+import { Model } from "sequelize";
+type GenericModel = Model<any, any>;
```
doodyparizada: why do you need this?

scripts/src/models/all.ts
```diff
@@ -1,83 +1,111 @@
...
 	blockchain_txid: Sequelize.STRING,
 	user_id: Sequelize.STRING, // redundent
 	meta: Sequelize.JSON, // some human readable data for UI
 	type: Sequelize.STRING,
 	value: Sequelize.JSON,
 	created_date: Sequelize.DATE,
-});
+}));
+
+models.sync().then(synced => console.log(`finished syncing all models (${ synced.length })`));
```
doodyparizada: I'd prefer everything related to sync to be in a separate file even if that creates the overhead of having to remember to add the models to the list of synced classes

scripts/src/routes/offers.ts
```diff
@@ -21,6 +21,10 @@ router.get("/", async (req, res, next) => {
 	}
 });
 
+router.get("/moo", async (req, res, next) => {
+	throw "moo";
```
doodyparizada: delete?

scripts/src/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+export enum IdPrefix {
+	User = "U",
+	App = "A",
+	Transaction = "T",
+	Offer = "O",
+	None = ""
+}
+
+const ID_LENGTH = 10;
+const ID_CHARS = "1234567890abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ".split("");
```
doodyparizada: Why does this have to be list? can't you access a string with `[]`?

scripts/src/models/all.ts
nitzantomer: Because they are using older version of typescript where you couldn't specify default generics constraints.  
So the definition for Model is:
```
interface Model<TInstance, TAttributes> extends Hooks<TInstance>, Associations { ... }
```
So you **have** to specify the generic constraints, and so instead of keep writing: `Model<any, any>` everywhere I created this type alias.

Now days you can do:
```
interface Model<TInstance = any, TAttributes = any> extends Hooks<TInstance>, Associations { ... }
```

scripts/src/models/all.ts
nitzantomer: it's a start. we'll obviously change things.  It was just a simple way to make sure that no model is left un-synced.

scripts/src/routes/offers.ts
nitzantomer: right.

scripts/src/utils.ts
nitzantomer: that's a good option.

scripts/src/models/all.ts
```diff
@@ -1,83 +1,110 @@
...
+	};
+}
+
+if (dbConfig.options.storage && !/^[./]/.test(dbConfig.options.storage)) {
+	dbConfig.options.storage = path(dbConfig.options.storage);
+}
+
+const sequelize = new Sequelize(dbConfig.database, dbConfig.username, dbConfig.passive, dbConfig.options);
+const models = {
+	_items: [] as Model<any, any>[],
```
doodyparizada: `GenericModel`

scripts/src/models/all.ts
```diff
@@ -1,83 +1,110 @@
...
+	push: function<T extends Model<any, any>>(item: T): T {
```
doodyparizada: `GenericModel`

scripts/src/models/all.ts
doodyparizada: @liorama said to put int npm scripts

PR #5, opened 2018-02-07
add user registration flow

doodyparizada:
user registration flow.

There are some open questions about when does the user get the 6000 get started KIN.

### DIFF
--- a/scripts/src/routes/users.ts
+++ b/scripts/src/routes/users.ts
@@ -1,0 +1,56 @@
+import { Router } from "express";
+import {AuthToken, validateJWT} from "../services/users";
+import * as db from "../models/all";
+
+export const router: Router = Router();
+
+export type AuthToken = {
+	token: string;
+	activated: boolean;
+}
+
+/**
+ * sign in a user
+ */
+router.post("/", async (req, res, next) => {
+	req.body.jwt;
+	req.body.device_id;
+	req.body.public_address;
+	let {userId , appId} = validateJWT(req.body.jwt); // throws if JWT not valid // XXX test this case
+
+	let user = db.Users.find({appId: appId, appUserId: userId});
+	if (!user) {
+		// new user
+		user = db.User.create(appId, userId, req.body.public_address, /*activated = */ false);
+		// create wallet with lumens:
+		// kin.sdk.createWallet(req.body.public_address);
+	}
+
+	let authToken = db.AuthToken.create(user.id, req.body.device_id);
+
+	res.status(200).send({token: authToken.token, activated: user.activated});
+});
+
+/**
+ * user activates by approving TOS
+ */
+router.post("/me/activate", async (req, res, next) => {
+	const result = null; // activate user if not already activated and fund the account with KIN 
+
+	let token = req.token;
+	let userId = db.AuthToken.find(req.token);
+	let user = db.Users.find(userId);
+	let deviceId = db.AuthToken.get(token).deviceId;
+
+	if (!user.activated) {
+		user.activated = true;
+		token = db.AuthToken.create(user.id, deviceId); // we want the auth token to be some sort of encoded payload so db access isn't needed to authenticate
+		// create an order for Getting Started Invisible Offer
+		// order = Offers.createOrder(GETTING_STARTED_OFFER_ID)
+		// order.submitEarn(); which does:
+		// tx_id = kin.sdk.payTo(public_address, order.id);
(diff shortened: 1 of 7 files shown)


scripts/src/routes/users.ts
```diff
@@ -0,0 +1,56 @@
...
+export type AuthToken = {
+	token: string;
+	activated: boolean;
+}
+
+/**
+ * sign in a user
+ */
+router.post("/", async (req, res, next) => {
+	req.body.jwt;
```
nitzantomer: what's up with these 3 lines?

scripts/src/routes/users.ts
```diff
@@ -0,0 +1,56 @@
...
+	req.body.device_id;
+	req.body.public_address;
+	let {userId , appId} = validateJWT(req.body.jwt); // throws if JWT not valid // XXX test this case
```
nitzantomer: can be `const`

scripts/src/routes/users.ts
```diff
@@ -0,0 +1,56 @@
...
+
+	let user = db.Users.find({appId: appId, appUserId: userId});
```
nitzantomer: can be `const`

scripts/src/routes/users.ts
```diff
@@ -0,0 +1,56 @@
...
+	if (!user) {
+		// new user
+		user = db.User.create(appId, userId, req.body.public_address, /*activated = */ false);
+		// create wallet with lumens:
+		// kin.sdk.createWallet(req.body.public_address);
+	}
+
+	let authToken = db.AuthToken.create(user.id, req.body.device_id);
```
nitzantomer: can be `const`

scripts/src/routes/users.ts
```diff
@@ -0,0 +1,56 @@
...
+
+	res.status(200).send({token: authToken.token, activated: user.activated});
+});
+
+/**
+ * user activates by approving TOS
+ */
+router.post("/me/activate", async (req, res, next) => {
+	const result = null; // activate user if not already activated and fund the account with KIN
```
nitzantomer: doesn't seem like you're using this variable.

scripts/src/services/users.ts
```diff
@@ -0,0 +1,54 @@
...
+		key: string;
+	};
+	payload: JWTClaims & {
+		// custom claims
+		user_id: string;
+	};
+	signature: string;
+}
+
+function getApplicationPublicKey(application_id: string, key_id: string) {
```
nitzantomer: camel case:
```
function getApplicationPublicKey(applicationId: string, keyId: string) {
```

scripts/src/services/users.ts
```diff
@@ -0,0 +1,54 @@
...
+	// return the public key for the given application.
+	// an application might have multiple keys. each key identified by key_id.
+
+	const publicKeys = {
```
nitzantomer: I think it's better to put this in another file and load (import) it

scripts/src/services/users.ts
```diff
@@ -0,0 +1,54 @@
...
+		kik: {one: "-----BEGIN PUBLIC KEY-----\n" +
+			"MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDdlatRjRjogo3WojgGHFHYLugdUWAY9iR3fy4arWNA1KoS8kVw33cJibXr8bvwUAUparCwlvdbH6dvEOfou0/gCFQsHUfQrSDv+MuSUMAe8jzKE4qW+jK+xQU9a03GUnKHkkle+Q0pX/g6jXZ7r1/xAK5Do2kQ+X5xK9cipRgEKwIDAQAB\n" +
+			"-----END PUBLIC KEY-----"}};
+
+	return publicKeys[application_id][key_id];
+}
+
+export function validateJWT(jwt: string): {userId: string, appId: string} {
+
+	/*
```
nitzantomer: what's this comment?

scripts/src/services/users.ts
```diff
@@ -0,0 +1,54 @@
...
+	jsonwebtoken.verify(jwt, publicKey);
+
+	return {
+		userId: decoded.payload.user_id,
+		appId: decoded.payload.iss,
+	};
+}
+
+function getUser(userId: string) {
+	return;
```
nitzantomer: nice

scripts/src/services/users.ts
doodyparizada: for sure - it will be loaded from the DB and cached in memory

doodyparizada:
I merging this as is. We are migrating to a new ORM which will change most of this code.

PR #7, opened 2018-02-08
Code style

nitzantomer:


code_style_scheme.xml
```diff
@@ -0,0 +1,80 @@
...
+    </indentOptions>
+  </codeStyleSettings>
+  <codeStyleSettings language="SCSS">
+    <indentOptions>
+      <option name="INDENT_SIZE" value="4" />
+      <option name="USE_TAB_CHARACTER" value="true" />
+      <option name="SMART_TABS" value="true" />
+    </indentOptions>
+  </codeStyleSettings>
+  <codeStyleSettings language="Stylus">
```
doodyparizada: Why do we need all these? why not just javascript and typescript?

package.json
```diff
@@ -19,9 +23,11 @@
     "@types/winston": "^2.3.7",
     "npm-run-all": "^4.0.1",
     "rimraf": "2.6.1",
+    "sqlite3": "3.1.13",
     "swagger-node-codegen": "^1.1.0",
-    "typescript": "^2.6.2",
-    "sqlite3": "3.1.13"
+    "tslint": "^5.9.1",
```
doodyparizada: Can you add this to the pre-test phase in travis?

scripts/src/config.ts
```diff
@@ -1,19 +1,20 @@
...
+export interface DatabaseConfig {
 	database: string;
 	username: string;
 	password: string;
 	options: SqlOptions;
-};
+}
 
-export type Config = {
+export interface Config {
```
doodyparizada: why change from type?

scripts/src/logging.ts
```diff
@@ -31,11 +31,11 @@ type WinstonTransportOptions = GenericTransportOptions & GenericTextTransportOpt
 function createTarget(target: LogTarget): winston.TransportInstance {
 	let cls: { new (options: WinstonTransportOptions): winston.TransportInstance };
 	const defaults: WinstonTransportOptions = {
-		timestamp: true
+		timestamp: true,
```
doodyparizada: 👍

scripts/src/models/all.ts
```diff
@@ -25,86 +24,86 @@ if (dbConfig.options.storage && !/^[./]/.test(dbConfig.options.storage)) {
...
 const User = models.push(sequelize.define("users", {
-	id: { type: Sequelize.STRING, primaryKey: true }, //internal id
+	activated_date: Sequelize.DATE,
 	app_id: Sequelize.STRING,
 	app_user_id: Sequelize.STRING,
 	created_date: Sequelize.DATE,
-	activated_date: Sequelize.DATE
-}));
+	id: { type: Sequelize.STRING, primaryKey: true }, // internal id
+}) as GenericModel);
```
doodyparizada: I condemn this.
:(

scripts/src/routes/offers.ts
```diff
@@ -15,8 +16,8 @@ router.get("/", async (req, res, next) => {
 		res.status(result.code || 200).send(result.data);
 	} catch (err) {
 		return res.status(500).send({
+			error: "Server Error",
```
doodyparizada: don't invest on these 500 errors. We should have a generic error handler that knows how to serialize exceptions thrown from the code.

scripts/src/config.ts
nitzantomer: the linter prefers interfaces over type aliases ([interface-over-type-literal](https://palantir.github.io/tslint/rules/interface-over-type-literal/)), because of [this part of the docs](https://www.typescriptlang.org/docs/handbook/advanced-types.html):

> A second more important difference is that type aliases cannot be extended or implemented from (nor can they extend/implement other types). Because an ideal property of software is being open to extension, **you should always use an interface over a type alias if possible**.

scripts/src/models/all.ts
nitzantomer: yeah me too. but it's just to get this linter in.
as we've already agreed, this part is going to look completely different.

scripts/src/routes/offers.ts
nitzantomer: yeah, just fixing linting errors...

code_style_scheme.xml
nitzantomer: it's what webstorm created when i exported the settings

PR #9, opened 2018-02-12
switched from sequelize to typeorm

doodyparizada:


config/default.json
```diff
@@ -6,19 +6,13 @@
...
-			"operatorsAliases": false,
-			"pool": {
-				"max": 5,
-				"min": 0,
-				"acquire": 30000,
-				"idle": 10000
-			}
-		}
+		"database": "database.sqlite",
+		"synchronize": true,
```
doodyparizada: I want `synchronize: false` and to have a separate runnable file that knows how to generate tables.

scripts/src/index.ts
```diff
@@ -42,6 +47,9 @@ app.use((err, req, res, next) => {
 	res.status(status).send({ status, error: "Server error" });
 });
 
+// initializing db and models
+initModels().then(str => logger.debug(str));
```
doodyparizada: msg?

scripts/src/models/index.ts
```diff
@@ -0,0 +1,47 @@
+import "reflect-metadata";
+import { BaseEntity, createConnection } from "typeorm";
+import { ConnectionOptions } from "typeorm/connection/ConnectionOptions";
+
+import { getConfig } from "../config";
+import { normalizeError, path } from "../utils";
+
+const dbConfig = Object.assign(getConfig().db);
+if (dbConfig.type === "sqlite" && !/^[./]/.test(dbConfig.database)) {
```
doodyparizada: what?

scripts/src/models/index.ts
```diff
@@ -0,0 +1,47 @@
...
+		.then(connection => {
+			return createOnConnectedString(connection.options);
+		})
+		.catch(error => {
+			return normalizeError(error);
+		});
+}
+
+function createOnConnectedString(options: ConnectionOptions): string {
+	let str = `connected to ${ options.type } server`;
```
doodyparizada: msg

scripts/src/models/offers.ts
```diff
@@ -0,0 +1,185 @@
+import { Column, Entity, PrimaryColumn } from "typeorm";
+
+import { Model, Register } from "./index";
+
+export type OfferMeta = {
+	title: string;
+	image: string;
+	description: string;
+};
```
doodyparizada: 👍

scripts/src/models/offers.ts
```diff
@@ -0,0 +1,185 @@
...
+	private _amount: number;
+
+	@Column({ name: "cap" })
+	private _cap: string;
+
+	@Column({ name: "meta" })
+	private _meta: string;
+
+	@Column({ name: "type" })
+	private _type: "spend" | "earn";
```
doodyparizada: this enum should be defined outside and used by all places that need it

scripts/src/models/offers.ts
```diff
@@ -0,0 +1,185 @@
...
+	public get createdDate(): Date {
+		return this._createdDate;
+	}
+
+	public get type(): "coupon" {
+		return this._type;
+	}
+
+	public get isUsed(): boolean {
+		return this._isUsed;
```
doodyparizada: I'm thinking that we may want usedBy

scripts/src/models/users.ts
```diff
@@ -0,0 +1,139 @@
...
+	private _appId: string;
+
+	@Column({ name: "app_user_id" })
+	private _appUserId: string;
+
+	@Column({ name: "created_date" })
+	private _createdDate: Date;
+
+	@Column({ name: "activated_date" })
+	private _activatedDate: Date;
```
doodyparizada: reminder: yohay suggested to have a permissions object that each token encapsulates and a user has to allow permissions per token.

scripts/src/models/users.ts
```diff
@@ -0,0 +1,139 @@
...
+
+	@Column({ name: "name" })
+	private _name: string;
+
+	@Column({ name: "jwt_public_key" })
+	private _jwtPublicKey: string;
+
+	constructor() {
+		super();
+	}
```
doodyparizada: Do you have to define a constructor if all it does is call super?

config/default.json
nitzantomer: This is part of the `default` config, which (at least for now) is the "dev config".
In a different config file you can set it to `false`, but for development it is easier with it as `true`

scripts/src/models/index.ts
nitzantomer: Well, if using sqlite and the db file path isn't relative/absolute then make the path relative to the root directory of the project.
For example, take what's in our `config/default.json` configuration:
```
...
"database": "database.sqlite",
...
```
The file that will be used will be `PROJECT_ROOT/database.sqlite`

scripts/src/models/offers.ts
nitzantomer: Maybe, so we'll add it when/if needed

doodyparizada:
I rebased from master. The main conflict is that I added default values to the models which we need to reimplement according to b68153d13485a082638804ee33336b1f34bb08ef

PR #11, opened 2018-02-13
model default values and using simple-json column type

nitzantomer:


scripts/src/models/index.ts
```diff
@@ -17,6 +17,11 @@ export abstract class Model extends BaseEntity {
 	@PrimaryColumn({ name: "id" })
 	private _id: string;
 
+	protected constructor(prefix: IdPrefix = IdPrefix.None) {
+		super();
+		this._id = generateId(prefix);
```
doodyparizada: just a behavioural check, we want the following to happen:
```typescript
let user: User = new User();
user.id === "UXXXXXX"; ?  // newly generated id
```

but, as an example, we don't want `Users.findOne()` to execute the constructor and create a new ID.

nitzantomer: Yeah, I'm aware of that.
My guess is that the orm will create an instance, which will have a newly generated id which will then be overridden with the values from the db.

It's just a guess, we'll test it once we start experimenting, and we'll change/optimize as we move forward.

PR #16, opened 2018-02-14
re implement user reg flow

doodyparizada:


scripts/src/services/users.ts
```diff
@@ -46,3 +55,63 @@ export function validateJWT(jwt: string): {userId: string, appId: string} {
...
+export async function getOrCreateUserCredentials(
+	jwt: string,
+	walletAddress: string,
+	deviceId: string): Promise<AuthToken> {
+
+	const { userId , appId } = validateJWT(jwt); // throws if JWT not valid // XXX test this case
+	let user = await db.User.findOne( { appId, appUserId: userId });
+	if (!user) {
+		// new user
+		user = new db.User();
```
nitzantomer: This is an ugly way of doing it.
You can do this:
```
user = Object.assign(new db.User(), {
    appId, userId, walletAddress
});
```
And should also place that in the `User` ctor

scripts/src/services/users.ts
```diff
@@ -46,3 +55,63 @@ export function validateJWT(jwt: string): {userId: string, appId: string} {
...
+export async function activateUser(token: string): Promise<AuthToken> {
+	let authToken = await db.AuthToken.findOne({ token });
+	const user = await db.User.findOneById( authToken.userId );
+
+	if (!user.activated) {
+		user.activatedDate = new Date();
+		// XXX use transaction
+		await user.save();
+
+		authToken = new db.AuthToken();
```
nitzantomer: Same as with the new user, do it in the ctor and user `Object.assign`

scripts/src/services/users.ts
```diff
@@ -46,3 +55,63 @@ export function validateJWT(jwt: string): {userId: string, appId: string} {
...
+		authToken.userId = user.id;
+		authToken.deviceId = authToken.deviceId;
+		// XXX should implement some sort of authtoken scoping that will be encoded into the token:
```
nitzantomer: Maybe use the multiline comment style? (`/*...*/`)

scripts/src/services/users.ts
doodyparizada: yes, It's a copy from their documentation. Usually I wrap this kind of creation in a static `create` function

PR #22, opened 2018-02-20
Doody/earn flow

doodyparizada:
open orders are in-memory objects - should be redis
added curl examples to examples dir
order.meta is directly filled from offer.meta - should be filled from offer.order_meta or similar


scripts/src/models/orders.ts
```diff
@@ -10,14 +10,20 @@ export type TransactionMeta = {
...
+	sender_address?: string;
+	recipient_address?: string;
+};
+
 @Entity({ name: "orders" })
 @Register
 export class Order extends CreationDateModel {
 	@Column()
-	public type: string;
+	public type: "earn" | "spend";
```
nitzantomer: there's a type alias for that in `offers.ts`:
```
export type OfferType = "spend" | "earn";
```

scripts/src/models/offers.ts
```diff
@@ -54,7 +54,7 @@ export class OfferContent extends Model {
 	public content: string;
 
 	@Column({ name: "content_type" })
-	public contentType: "poll";
+	public contentType: "poll" | "coupon";
```
nitzantomer: maybe turn it into a type alias?

scripts/src/models/orders.ts
```diff
@@ -28,17 +34,27 @@ export class Order extends CreationDateModel {
...
+	public amount: number;
+
 	public constructor() {
 		super(IdPrefix.Transaction);
 	}
 
-	public get status(): "open" | "done" | "failed" | "pending" {
-		if (this.blockchainTxId) {
-			return "done";
+	public get status(): "completed" | "failed" | "pending" {
```
nitzantomer: maybe this should be a type alias as well?

scripts/src/models/orders.ts
```diff
@@ -28,17 +34,27 @@ export class Order extends CreationDateModel {
...
+		if (this.blockchainData) {
+			return "completed";
 		}
-		return "open";
+		return "pending";
 	}
 }
+
+export class OpenOrder {
```
nitzantomer: why is this a class and not an interface/type alias?

scripts/src/services/orders.ts
```diff
@@ -67,127 +71,118 @@ export interface Order {
...
-			return order;
-		}
+	const order = await db.Order.findOneById(orderId);
+	if (!order) {
+		throw Error(`no such order ${orderId}`); // XXX throw and exception that is convert-able to json
+	}
+	return orderDbToApi(order);
+}
+
+const openOrdersDB: Map<string, db.OpenOrder> = new Map<string, db.OpenOrder>();
```
nitzantomer: No need for the type annotation, this should suffice:
```
const openOrdersDB = new Map<string, db.OpenOrder>();
```

PR #25, opened 2018-02-21
add orders to create script

doodyparizada:
This is to let the clients create a few users that have pre loaded data with all the states


scripts/src/create.ts
```diff
@@ -83,12 +83,61 @@ async function createOffers(): Promise<Offer[]> {
 	return offers;
 }
 
+function orderFromOffer(offer: Offer, userId: string): Order {
+	const order = new Order();
+	order.userId = userId;
+	order.offerId = offer.id;
+	order.meta = {
```
nitzantomer: This is nicer:
```
order.meta = Object.assign({}, offer.meta, { call_to_action: "press here" });
```

PR #28, opened 2018-02-26
Doody/offer contents

doodyparizada:


scripts/src/services/offer_contents.ts
```diff
@@ -0,0 +1,51 @@
...
+};
+
+export async function getOffer(offerId: string): Promise<db.OfferContent> {
+	return await db.OfferContent.findOne({ offerId });
+}
+
+export async function isValid(offerId: string, form: string): Promise<boolean> {
+	const parsed = JSON.parse(form);
+
+	const offer: db.OfferContent = await getOffer(offerId);
```
nitzantomer: no need to specify the type:
```
const offer = await getOffer(offerId);
```

scripts/src/services/offer_contents.ts
```diff
@@ -0,0 +1,51 @@
...
+	const poll: Poll = JSON.parse(offer.content);
+
+	for (const page of poll.pages) {
+		const qId: string = page.question.id;
```
nitzantomer: no need to specify the types:
```
const qId = page.question.id;
const ans = page.question.answers;
```

scripts/src/services/offer_contents.ts
```diff
@@ -0,0 +1,51 @@
...
+		},
+	}],
+};
+
+export async function getOffer(offerId: string): Promise<db.OfferContent> {
+	return await db.OfferContent.findOne({ offerId });
+}
+
+export async function isValid(offerId: string, form: string): Promise<boolean> {
+	const parsed = JSON.parse(form);
```
nitzantomer: what is the type of `parsed`?:
```
const parsed = JSON.parse(form) as ?;
```

scripts/src/services/offer_contents.ts
```diff
@@ -0,0 +1,51 @@
...
+
+	const offer: db.OfferContent = await getOffer(offerId);
+	const poll: Poll = JSON.parse(offer.content);
+
+	for (const page of poll.pages) {
+		const qId: string = page.question.id;
+		const ans: string[] = page.question.answers;
+
+		if (ans.includes(parsed[qId])) {
+			return false;
```
nitzantomer: isn't it the other way around?  return `true` here and `false` at the end of the function?

PR #37, opened 2018-03-12
add coupon info page

doodyparizada:


scripts/src/models/orders.ts
```diff
@@ -8,6 +8,7 @@ export type OrderMeta = {
 	title: string;
 	description: string;
 	call_to_action?: string;
+	content?: string;
```
nitzantomer: looks like the content always contains the same structure.
maybe it should be a typed value instead of a string?

scripts/src/services/orders.ts
```diff
@@ -33,7 +30,8 @@ export interface OpenOrder {
 
 export interface Order {
 	id: string;
-	result?: OrderResult;
+	result?: AssetValue | FailureReason;
+	content?: string; // json serialized payload of the coupon page
```
nitzantomer: typed value?

scripts/src/models/orders.ts
doodyparizada: this helps with encapsulation of this data across the APIs. I do agree that we can benefit from defining the data structure as an API in the ecosystem-api repo

PR #41, opened 2018-03-14
Doody/offer blockchain data

doodyparizada:
added the option to submit a spend order + add recipient_address to offer blockchain data.
currently there is no integration with blockchain for spend offer, so on client submission the server simulates the payment.

### DIFF
--- a/scripts/src/create.ts
+++ b/scripts/src/create.ts
@@ -3,22 +3,23 @@
  * All the names of companies, products and KIN values are completely made up and are used for TESTING only.
  */
 import * as fs from "fs";
-import { User, AuthToken } from "./models/users";
+import { AuthToken, User } from "./models/users";
 import { Application } from "./models/applications";
-import { Offer, OfferContent, AppOffer, Asset, OfferOwner } from "./models/offers";
-import { Order } from "./models/orders";
+import { AppOffer, Asset, Offer, OfferContent, OfferOwner } from "./models/offers";
+import { OpenOrder, Order } from "./models/orders";
 
 import { init as initModels } from "./models";
 import { getConfig } from "./config";
 import {
 	animalPoll,
-	kikPoll,
-	tutorial,
-	Poll,
-	Tutorial,
 	CouponInfo,
-	CouponOrderContent
+	CouponOrderContent,
+	kikPoll,
+	Poll,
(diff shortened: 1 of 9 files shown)


scripts/src/create.ts
```diff
@@ -3,14 +3,23 @@
...
-import { animalPoll, kikPoll, tutorial, Poll, Tutorial, CouponInfo, CouponOrderContent } from "./services/offer_contents";
+import {
+	animalPoll,
+	CouponInfo,
+	CouponOrderContent,
+	kikPoll,
+	Poll,
+	tutorial,
+	Tutorial
+} from "./services/offer_contents";
```
liorama: I like it the way it was before, a one liner, it's consistent with our other imports.

scripts/src/mock_client.ts
```diff
@@ -170,7 +170,7 @@ async function earnTutorial() {
 	let earn: Offer;
 
 	for (const offer of offers.offers) {
-		if (offer.title === "Getting Started") {
+		if (offer.description === "Kin Tutorial") {
```
liorama: Use a const

scripts/src/services/orders.ts
```diff
@@ -80,54 +96,70 @@ export async function submitEarn(
...
-export async function cancelOrder(options, logger: LoggerInstance): Promise<void> {
-	return;
+export async function cancelOrder(orderId: string, logger: LoggerInstance): Promise<void> {
+	// you can only delete an open order - not a pending order
+	// validate order
+	const openOrder: db.OpenOrder = openOrdersDB.get(orderId);
+	if (!openOrder) {
+		throw Error(`no such order ${orderId}`);
+	}
+	openOrdersDB.delete(openOrder.id);
```
liorama: It'a minute but it's a bit better (best practice) to use `orderId`, they are obviously the same at this point and the compiler won't have to look up the property on the object.

scripts/src/create.ts
doodyparizada: The auto-formatter does this

scripts/src/services/orders.ts
doodyparizada: is this just an optimization or a javascript best practice?

PR #42, opened 2018-03-15
use Model.new instead of ctor

nitzantomer:


scripts/src/models/index.ts
```diff
@@ -14,24 +15,39 @@ let connection: Connection;
...
-		this.id = generateId(prefix);
+		return map;
 	}
+
+	@PrimaryColumn()
+	public id: string;
 }
 
 export abstract class CreationDateModel extends Model {
+	protected static initializers = Model.copyInitializers({ createdDate: () => new Date() });
```
doodyparizada: can a Column accept a default value as a function instead?

scripts/src/create.ts
```diff
@@ -236,18 +236,36 @@ async function createApp(appId, apiKey, name) {
 	const jwtPublic = fs.readFileSync("./examples/jwt_public_key.pem", "utf-8");
```
doodyparizada: here `orderFromOffer` there is `new Order()`

scripts/src/models/index.ts
```diff
@@ -3,37 +3,77 @@ import { BaseEntity, Column, createConnection, PrimaryColumn, Connection, Connec
...
-		super(prefix);
-		this.createdDate = new Date();
-	}
 }
 
 export function register(ctor: ModelConstructor) {
 	entities.push(ctor);
 }
 
+export function initializer(propName: string, fn: ModelMemberInitializer) {
```
doodyparizada: `return initializers({[propName]: fn})`

scripts/src/public/services/orders.ts
```diff
@@ -106,7 +106,15 @@ export async function submitOrder(
 	}
 
 	// transition open order to pending order
-	const order = new db.Order(openOrder, offer);
+	const order = db.Order.new({
+		id: openOrder.id,
+		userId: openOrder.userId,
+		offerId: openOrder.offerId,
+		amount: offer.amount,
+		type: "earn",
```
doodyparizada: `		type: offer.type,`

scripts/src/models/index.ts
nitzantomer: 👍

PR #48, opened 2018-03-18
Spend flow test

doodyparizada:
add a test for spend flow


scripts/src/internal/app.ts
```diff
@@ -30,3 +34,11 @@ createRoutes(app);
 app.use(notFoundHandler);
 // catch errors
 app.use(generalErrorHandler);
+
+// initializing db and models
+initModels().then(msg => {
+	logger.debug("init db", { msg });
```
nitzantomer: why `, { msg }` and not just `, msg`?

scripts/src/internal/app.ts
```diff
@@ -30,3 +34,11 @@ createRoutes(app);
...
+	initPaymentCallbacks(logger).then(res => {
+		logger.debug("init payment result", { res });
```
nitzantomer: same here with the `, { res }`

scripts/src/internal/services.ts
```diff
@@ -69,3 +71,15 @@ export async function paymentFailed(payment: CompletedPayment, reason: string, l
...
 	logger.info(`failed order with payment <${payment.id}, ${payment.transaction_id}>`);
 }
+
+/**
+ * register to get callbacks for incoming payments for all the active offers
+ */
+export async function initPaymentCallbacks(logger: LoggerInstance): Promise<Watcher> {
+	const offers = await Offer.find({ type: "spend" }); // get all active spend offers
+	// create a list of unique addresses
+	const addresses: string[] = removeDups(offers.map(offer => offer.blockchainData.recipient_address));
```
nitzantomer: no need for the type annotation here, can be: 
```const addresses = removeDups(offers.map(offer => offer.blockchainData.recipient_address));```

File: scripts/src/mock_client.ts [hunk 4]
```diff
@@ -31,13 +32,19 @@ const STELLAR = new Stellar("testnet");
 class Client {
 
 	public authToken: AuthToken;
-	private keyPair: StellarSdk.KeyPair;
+	private keyPair: StellarSdk.Keypair = null;
```
nitzantomer: why is it important to intialize it to `null`?

scripts/src/mock_client.ts
```diff
@@ -46,25 +53,40 @@ class Client {
...
 
-	public async pay(recipientAddress: string, amount: number): Promise<StellarSdk.TransactionResult> {
+	public async pay(recipientAddress: string, amount: number, orderId: string): Promise<TransactionRecord> {
 		const op = StellarSdk.Operation.payment({
 			destination: recipientAddress,
 			asset: STELLAR.kinAsset,
-			amount
+			amount: amount.toString()
 		});
+		const memoText = `1-${this.appId}-${orderId}`;
```
nitzantomer: extract the version part to a const of something

scripts/src/utils.ts
```diff
@@ -64,3 +64,11 @@ export function pick<T, K extends keyof T>(obj: T, ...props: K[]): Pick<T, K> {
 	props.forEach(name => newObj[name] = obj[name]);
 	return newObj;
 }
+
+export function removeDups(arr: string[]): string[] {
```
nitzantomer: better implementation:
```
export function removeDuplicates(arr: string[]): string[] {
    return Array.from(new Set(arr));
}
```

scripts/src/internal/app.ts
doodyparizada: so it shows up in the log as msg=msg

scripts/src/internal/app.ts
doodyparizada: same answer

PR #50, opened 2018-03-21
Nitzan/strict mode

nitzantomer:
Issue: https://kinecosystem.atlassian.net/projects/ECO/issues/ECO-139

### DIFF
--- a/scripts/src/config.ts
+++ b/scripts/src/config.ts
@@ -17,6 +17,6 @@ export function getConfig<T extends Config>(filePath?: string): T {
 		return config as T;
 	}
 
-	config = require(path(filePath));
+	config = require(path(filePath!));
 	return config as T;
 }
--- a/scripts/src/config.ts
+++ b/scripts/src/config.ts
@@ -12,11 +12,11 @@ export interface Config {
 }
 
 let config: Config;
-export function getConfig<T extends Config>(filePath?: string): T {
-	if (config) {
-		return config as T;
-	}
 
+export function init(filePath: string) {
 	config = require(path(filePath!));
+}
+
+export function getConfig<T extends Config>(): T {
 	return config as T;
 }
--- a/scripts/src/internal/index.ts
+++ b/scripts/src/internal/index.ts
@@ -4,7 +4,7 @@ import { initLogger } from "../logging";
 
 import { getConfig } from "./config";
 const config = getConfig();
-const logger = initLogger(...config.loggers);
+const logger = initLogger(...config.loggers!);
 
 import { app } from "./app";
 
@@ -16,7 +16,7 @@ server.on("listening", onListening);
 /**
  * Event listener for HTTP server "error" event.
  */
-function onError(error) {
+function onError(error: Error & { syscall: string; code: string; }) {
 	if (error.syscall !== "listen") {
 		throw error;
 	}
--- a/scripts/src/internal/index.ts
+++ b/scripts/src/internal/index.ts
@@ -4,6 +4,7 @@ import { ServerError } from "../utils";
 import { initLogger } from "../logging";
 
 import { getConfig } from "./config";
+
 const config = getConfig();
 const logger = initLogger(...config.loggers!);
 
--- a/scripts/src/internal/index.ts
+++ b/scripts/src/internal/index.ts
@@ -1,5 +1,6 @@
 import * as http from "http";
 
+import { ServerError } from "../utils";
 import { initLogger } from "../logging";
 
 import { getConfig } from "./config";
@@ -16,7 +17,7 @@ server.on("listening", onListening);
 /**
  * Event listener for HTTP server "error" event.
  */
-function onError(error: Error & { syscall: string; code: string; }) {
+function onError(error: ServerError) {
 	if (error.syscall !== "listen") {
 		throw error;
 	}
--- a/scripts/src/internal/routes.ts
+++ b/scripts/src/internal/routes.ts
@@ -1,4 +1,4 @@
-import { Request, Response, Router, Express } from "express";
+import { Request, Response, Router, Express, RequestHandler } from "express";
 
 import {
 	CompletedPayment,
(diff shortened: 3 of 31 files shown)


scripts/src/internal/routes.ts
```diff
@@ -1,20 +1,20 @@
...
 	paymentComplete as paymentCompleteService,
 	paymentFailed as paymentFailedService,
 } from "./services";
 
-export async function paymentComplete(req: Request, res) {
+export const paymentComplete = async function(req: Request, res: Response) {
 	await paymentCompleteService(req.body as CompletedPayment, req.logger);
 	res.status(200).send({ status: "ok" });
-}
+} as any as RequestHandler;
```
nitzantomer: this is ugly but needed.
the problem is that the compiler has issues with it because we're not specifying the last (`next`) param.

doodyparizada:
tests are failing

scripts/src/config.ts
```diff
@@ -17,6 +17,6 @@ export function getConfig<T extends Config>(filePath?: string): T {
 		return config as T;
 	}
 
-	config = require(path(filePath));
```
doodyparizada: why not separate to `initConfig(filePath: string)` and `getConfig()`?

scripts/src/internal/index.ts
```diff
@@ -4,7 +4,7 @@ import { initLogger } from "../logging";
 
 import { getConfig } from "./config";
 const config = getConfig();
-const logger = initLogger(...config.loggers);
+const logger = initLogger(...config.loggers!);
```
doodyparizada: why init logger twice? in index and app?

scripts/src/internal/routes.ts
doodyparizada: so lets define the next parameter and just not use it. It's better than this

scripts/src/logging.ts
```diff
@@ -77,11 +77,13 @@ function createTarget(target: LogTarget): winston.TransportInstance {
...
 }
 
+type OptionsKey = keyof WinstonTransportOptions;
 function mergeOptions(defaults: WinstonTransportOptions, options: WinstonTransportOptions): WinstonTransportOptions {
 	const result = Object.assign({}, defaults);
+
 	Object.keys(options)
-		.filter(key => options[key] !== null && options[key] !== undefined)
-		.forEach(key => result[key] = options[key]);
+		.filter(key => options[key as OptionsKey] !== null && options[key as OptionsKey] !== undefined)
```
doodyparizada: `.filter(key: OptionsKey => options[key] !== null && options[key] !== undefined`

scripts/src/middleware.ts
```diff
@@ -33,7 +34,7 @@ export function requestLogger(req: express.Request, res, next) {
 						args = [...args, { reqId: id }];
 					}
 
-					target[name](...args);
+					(target[name] as any)(...args);
```
doodyparizada: why is `as any` needed here?

scripts/src/middleware.ts
```diff
@@ -45,9 +46,9 @@ export function requestLogger(req: express.Request, res, next) {
 	(req as any).id = id;
 	(req as any).logger = proxy;
 	next();
-}
+} as express.RequestHandler;
```
doodyparizada: why `as express.RequestHandler;`?

scripts/src/mock_client.ts
```diff
@@ -30,10 +30,9 @@ class Stellar {
 const STELLAR = new Stellar("testnet");
 
 class Client {
-
-	public authToken: AuthToken;
-	private keyPair: StellarSdk.Keypair;
-	private appId: string;
+	public authToken!: AuthToken;
+	private keyPair!: StellarSdk.Keypair;
+	private appId!: string;
```
doodyparizada: Why do we have to add `!` to all the definitions? it's ugly :(

scripts/src/mock_client.ts
```diff
@@ -199,12 +198,13 @@ class Client {
 	private async establishTrustLine(): Promise<TransactionRecord> {
 		const op = StellarSdk.Operation.changeTrust({
 			asset: STELLAR.kinAsset,
-			limit: undefined // XXX BUG this should be optional
+			limit: "" // XXX BUG this should be optional
```
doodyparizada: this will break the test. it HAS to be `undefined`

scripts/src/mock_client.ts
```diff
@@ -199,12 +198,13 @@ class Client {
...
 		});
 
+		let result: TransactionRecord;
 		for (let i = 0; i < 3; i++) {
 			try {
-				return await this.stellarOperation(op);
+				result = await this.stellarOperation(op);
```
doodyparizada: why? I don't like it

scripts/src/models/index.ts
```diff
@@ -13,12 +14,12 @@ let initPromise: Promise<string>;
 export type ModelConstructor = { new(): Model };
 export type ModelMemberInitializer = () => any;
 export abstract class Model extends BaseEntity {
-	public static new<T extends Model>(this: ObjectType<T>, data?: Partial<T>): T {
-		const instance = (this as typeof BaseEntity).create(data) as T;
+	public static new<T extends Model>(this: ObjectType<T>, data?: DeepPartial<T>): T {
+		const instance = (this as typeof BaseEntity).create(data!) as T;
 
 		for (const [name, initializer] of (this as typeof Model).initializers.entries()) {
```
doodyparizada: why not `name: keyof Model`

scripts/src/public/index.ts
```diff
@@ -16,7 +16,7 @@ server.on("listening", onListening);
 /**
  * Event listener for HTTP server "error" event.
  */
-function onError(error) {
+function onError(error: Error & { syscall: string; code: string; }) {
```
doodyparizada: Maybe define a new Type that encapsulates this because you're using it twice

scripts/src/public/services/offers.ts
```diff
@@ -37,20 +37,28 @@ export async function getOffers(
...
-				description: offer.meta.description,
-				image: offer.meta.image,
-				amount: offer.amount,
-				blockchain_data: offer.blockchainData,
-				offer_type: offer.type,
-				content: content.content,
-				content_type: content.contentType,
-			};
-		}));
+	const offers = await Promise.all(
```
doodyparizada: why not `const offers: Offer[] = ...`

doodyparizada:
@nitzantomer see the type orm example from nestjs that you recommended: https://docs.nestjs.com/recipes/sql-typeorm#. It doesn't have all these `!`

scripts/src/internal/routes.ts
nitzantomer: no, it's not better.  this "hack" is ts only, but the js code will be clean.
adding the `next` param will cause it to also be there in the js code.

scripts/src/models/index.ts
nitzantomer: there's no way to add type annotation when destructuring:
https://github.com/Microsoft/TypeScript/issues/7576

scripts/src/public/services/offers.ts
nitzantomer: because the type of the result is `Offer | undefined`, the compiler doesn't know that we're filtering the null values.
so we need to cast it (`as Offer[]`)

PR #51, opened 2018-03-26
add thank you page type

liorama:


scripts/src/public/services/offer_contents.ts
```diff
@@ -10,14 +10,15 @@ export interface Question {
...
+	"EarnThankYou",
 }
 
 export interface Poll {
 	pages: Array<{ type: PageType, title: string, description: string, question: Question }>;
 }
 
 export interface Tutorial {
-	pages: Array<{ type: PageType.ImageAndText, title: string, image: string, bodyHtml: string, footerHtml: string, buttonText: string }>;
+	pages: Array<{ type: PageType.ImageAndText | PageType.EarnThankYou, title?: string, image?: string, bodyHtml?: string, footerHtml?: string, buttonText?: string }>;
```
nitzantomer: use line breaks here:
```
    pages: Array<{ 
        type: PageType.ImageAndText | PageType.EarnThankYou;
        title?: string;
        ...
    }
```

Or even put that into a new type and then reference it.
Also, when writing types, please use `;` and not `,`

PR #52, opened 2018-03-26
ECO-165 Updated poll content

liorama:


scripts/src/create.ts
```diff
@@ -91,21 +92,15 @@ async function createOffers(): Promise<Offer[]> {
...
-		assetsBase + "earn_offer4.png", 30, "McDonald's", "Completed Poll",
-		animalPoll));
-	offers.push(await createEarn("Nike", "Run or walk?", "Answer a poll",
-		assetsBase + "earn_offer5.png", 30, "Nike", "Completed Poll",
-		animalPoll));
+	offers.push(await createEarn("Kik", "Tell us more", "Answer a poll",
+		assetsBase + "earn_offer2.png", 30, "Kik", "Completed Poll",
+		kikPoll));
+	offers.push(await createEarn("Kik", "Tell us more", "Answer a poll",
+		assetsBase + "earn_offer2.png", 30, "Kik", "Completed Poll",
```
doodyparizada: same image?

scripts/src/public/services/offer_contents.ts
```diff
@@ -37,15 +38,111 @@ export const kikPoll: Poll = {
...
 		},
-	}],
+	}, {
+		type: PageType.FullPageMultiChoice,
+		title: "What do you like to use the most when chatting with others on Kik?",
+		description: `Complete the poll to earn <span style='color:#047cfc;'>4,000</span> Kin`,
+		question: {
+			id: "what324",
+			answers: [
+				"Video chat", "Stickers", "Gifs", "Emojis", "Kik custom emojis", "Just regular text messages",
```
doodyparizada: When are we changing the answers to `id, value`?

scripts/src/public/services/offer_contents.ts
```diff
@@ -37,15 +38,111 @@ export const kikPoll: Poll = {
...
+			id: "how432565",
+			answers: [
+				"Friends/Family", "Social Media", "Website/Search Engine", "Ads",
+			],
+		},
+	},
+	],
+};
+
+export const kikPoll2: Poll = {
```
doodyparizada: maybe this is kinPoll?

scripts/src/public/services/offer_contents.ts
liorama: I know, We need to do it.

PR #54, opened 2018-03-26
refactor poll/tutorial types

liorama:


scripts/src/public/services/offer_contents.ts
```diff
@@ -13,19 +13,32 @@ export enum PageType {
...
+}
+
 export interface Poll {
-	pages: Array<{
-		type: PageType;
-		title: string;
-		description: string;
-		question: Question;
-	}>;
+	pages: Array<PollPage | EarnThankYouPage>;
```
nitzantomer: if you're referencing `EarnThankYouPage`, maybe move the definition to be above this one?

doodyparizada:
@liorama you didn't wait for the build to finish and this breaks the build

PR #58, opened 2018-03-29
Nitzan/eco 141/offer caps

nitzantomer:


scripts/src/public/services/offers.ts
```diff
@@ -46,10 +49,20 @@ function offerDbToApi(offer: db.Offer, content: db.OfferContent) {
 	};
 }
 
-async function filterOffers(offers: db.Offer[], logger: LoggerInstance): Promise<Offer[]> {
+async function filterOffers(userId: string, offers: db.Offer[], logger: LoggerInstance): Promise<Offer[]> {
 	return await Promise.all(
 		offers
 			.map(async offer => {
+				const total = await dbOrder.Order.count({ where: { offerId: offer.id } });
+				if (total === offer.cap.total) {
```
doodyparizada: I would check >=

scripts/src/public/services/offers.ts
```diff
@@ -46,10 +49,20 @@ function offerDbToApi(offer: db.Offer, content: db.OfferContent) {
...
+					return null;
+				}
+
+				const forUser = await dbOrder.Order.count({ where: { offerId: offer.id, userId } });
+				if (forUser === offer.cap.per_user) {
```
doodyparizada: I would check >=

scripts/src/public/services/orders.ts
```diff
@@ -63,103 +69,114 @@ function orderDbToApi(order: db.Order, logger: LoggerInstance): Order {
...
+			const forUser = await db.Order.count({
+				where: {
+					userId,
+					offerId
+				}
+			});
+
+			if (forUser === offer.cap.per_user) {
+				return undefined;
+			}
```
doodyparizada: This logic can be shared with getOffers

doodyparizada: plus, in getOffers you return `null` and here you return `undefined`

scripts/src/public/services/orders.ts
```diff
@@ -63,103 +69,114 @@ function orderDbToApi(order: db.Order, logger: LoggerInstance): Order {
...
+				type: offer.type,
+				status: "opened",
+				meta: offer.meta.order_meta
+			});
+			await order.save();
+			return order;
+		};
+
+		// order = await lock(createOrderResourceId, fn());
+		order = await fn();
```
doodyparizada: if you're putting this in a function, give it a name?

scripts/src/public/services/orders.ts
```diff
@@ -63,103 +69,114 @@ function orderDbToApi(order: db.Order, logger: LoggerInstance): Order {
...
 	};
 }
 
 export async function submitOrder(
 	orderId: string, form: string | undefined, walletAddress: string, appId: string, logger: LoggerInstance): Promise<Order> {
 
-	const openOrder = await db.OpenOrder.findOneById(orderId);
-	if (!openOrder) {
-		throw Error(`no such order ${orderId}`);
+	const order = await db.Order.findOneById(orderId);
```
doodyparizada: `db.Order.findOne({id: orderId, status: "opened"});`

scripts/src/public/services/orders.ts
```diff
@@ -63,103 +69,114 @@ function orderDbToApi(order: db.Order, logger: LoggerInstance): Order {
...
-	// setTimeout(makeFailed, order.expiration, order.id);
-	return;
+	metrics.submitOrder(offer.type, offer.id);
+	return orderDbToApi(order);
 }
 
 export async function cancelOrder(orderId: string, logger: LoggerInstance): Promise<void> {
 	// you can only delete an open order - not a pending order
-	// validate order
+	const order = await db.Order.findOneById(orderId);
```
doodyparizada: you can only cancel opened:
`db.Order.findOne({id: orderId, status: "opened"});`

scripts/src/public/services/orders.ts
```diff
@@ -63,103 +69,114 @@ function orderDbToApi(order: db.Order, logger: LoggerInstance): Order {
...
+	if (!order) {
+		throw Error(`no such order ${ orderId }`);
+	}
 
-	const openOrder = await db.OpenOrder.findOneById(orderId);
-	if (!openOrder) {
-		throw Error(`no such order ${orderId}`);
+	if (order.status !== "opened") {
```
doodyparizada: instead of this check, move it to the query

scripts/src/public/services/orders.ts
nitzantomer: in both cases what i wrote is temporary, we'll need to change it as it's not optimal AT ALL.

PR #66, opened 2018-04-12
ECO-247 remove sign-in data duplication

liorama:


scripts/src/public/routes/users.ts
```diff
@@ -18,14 +18,20 @@ export const getUser = async function(req: Request, res: Response) {
 	res.status(200).send({ user });
 } as any as RequestHandler;
 
-type SignInData = {
-	sign_in_type: "whitelist" | "jwt";
+type JwtSignInData = {
+	sign_in_type: "jwt";
+	device_id: string;
+	public_address: string;
+	jwt?: string;
```
nitzantomer: i'm pretty sure that in the case of jwt sign in, the `jwt` property is mandatory.

PR #68, opened 2018-04-14
test blockchain for earn tx + parse stellar errors

doodyparizada:
* now when stellar transaction fails the error is caught parsed and re-thrown
* added a function call for externalOrder
* added polling on blockchain for the orderId in earn flow


scripts/src/mock_client.ts
```diff
@@ -28,6 +38,7 @@ class Stellar {
 
 const STELLAR = new Stellar("testnet");
 
+
```
nitzantomer: pourquoi???

PR #70, opened 2018-04-16
api changes around registartion and open order

doodyparizada:


scripts/src/mock_client.ts
```diff
@@ -18,10 +18,12 @@ import {
 	CollectionPage
 } from "stellar-sdk";
 import { CompletedPayment } from "./internal/services";
+import { SpendPayloadOffer } from "./public/services/applications";
 
-// const BASE = "http://localhost:3000";
-const BASE = "https://api.kinmarketplace.com"; // production - XXX get this from env var?
-const APP_ID = "smpl";
+const BASE = "http://localhost:3000";
+const SAMPLE_BASE = "http://localhost:3002";
```
nitzantomer: maybe `JWT_SERVICE_BASE`

scripts/src/models/orders.ts
```diff
@@ -101,7 +101,7 @@ export class Order extends CreationDateModel {
 	public type!: OfferType;
 
 	@Column("simple-json", { name: "blockchain_data", nullable: true })
-	public blockchainData?: BlockchainData;
+	public blockchainData!: BlockchainData;
```
nitzantomer: why?
it can be nullable based on the column meta

scripts/src/public/jwt.ts
```diff
@@ -3,30 +3,36 @@ import * as jsonwebtoken from "jsonwebtoken";
 import { Application } from "../models/applications";
 
 export type JWTClaims = {
-	iss: string; // issuer
+	iss: string; // issuer - the app_id
 	exp: number; // expiration
 	iat: number; // issued at
 	sub: string; // subject
-};
+	};
```
nitzantomer: what's this indentation all about?

scripts/src/models/orders.ts
doodyparizada: what? now we always set the blockchain data on both marketplace and external orders

doodyparizada: same def as Offer table

doodyparizada:
@nitzantomer approved

PR #77, opened 2018-04-26
Nitzan/eco 173/routing

nitzantomer:


package.json
```diff
@@ -69,7 +69,7 @@
     "restart": "npm-run-all build start",
     "transpile-tests": "tsc -p tests",
     "test-system": "node scripts/bin/mock_client",
-    "test": "jest --forceExit"
+    "test": "jest --forceExit --runInBand"
```
doodyparizada: you branched out of the test fixing branch so this is a side effect

nitzantomer: once we merge the other PR this will go away, probably

PR #82, opened 2018-04-29
Nitzan/eco 174/error handling

nitzantomer:


doodyparizada:
you have some issues with the rebase. I would just cherry pick your changes on top of the master and re-push.

scripts/src/errors.ts
```diff
@@ -0,0 +1,129 @@
...
+	},
+	UnprocessableEntity: {
+		OpenedOrdersOnly: 1,
+		OpenedOrdersUnreturnable: 2
+	},
+	BadRequest: {
+		UnknownSignInType: 1,
+		WrongJWTAlgorithm: 2
+	}
+};
```
doodyparizada: Is these a way to make all these enums? Maybe even encode the final value to each type. i.e.: `NotFound.App: 4041`

nitzantomer: you can't have hierarchy with enums, and i started with encoding the final values, but then switched to the current version as i thought that the generic factory function should be responsible for the prefix code

scripts/src/errors.ts
```diff
@@ -0,0 +1,129 @@
...
+export function NoSuchPublicKey(appId: string, keyid: string) {
+	return NotFoundError(CODES.NotFound.App, `Key "${ keyid }" not found for iss "${ appId }"`);
+}
+
+function ConflictError(index: number, message: string) {
+	return new MarketplaceError(409, index, "Conflict", message);
+}
+
+export function ExternalOrderExhausted() {
+	return ConflictError(CODES.Conflict.ExternalOrderExhausted, "User already completed offer, or has a pending order");
```
liorama: This isn't a Conflict error, it's more of a resource gone, like a 410

scripts/src/errors.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+function LockedError(index: number, message: string) {
+	return new MarketplaceError(423, index, "Resource Locked", message);
+}
+
+export function OfferCapReached(id: string) {
+	return LockedError(CODES.Locked.OfferCapReached, `Cap reached for offer: ${ id }`);
```
liorama: 423 means the resource (File/Document/Object) is locked, there's no locking here but an exhausted resource this can be one of these:
1. 410 error because the offer is gone
2. a 403 since it's a limit on the user. 
3. a lot of services try to keep it simple and return a 404 (offer not found) but I don't think it's descriptive enough.

scripts/src/errors.ts
```diff
@@ -0,0 +1,129 @@
...
+function BadRequestError(index: number, message: string) {
+	return new MarketplaceError(400, index, "Bad Request", message);
+}
+
+export function UnknownSignInType(type: string) {
+	return BadRequestError(CODES.BadRequest.UnknownSignInType, `Unknown sign-in type: ${ type }`);
+}
+
+export function WrongJWTAlgorithm() {
+	return BadRequestError(CODES.BadRequest.UnknownSignInType, "only ES256 supported");
```
liorama: Doody should have used a const for the algo test and you should import it here and not have it hard-coded

scripts/src/errors.ts
nitzantomer: explain what?

scripts/src/errors.ts
doodyparizada: There could be a resource locked error when locking is implemented. I agree that here it should be something else, I tend to 410

doodyparizada:
@nitzantomer fix the weird commit history

scripts/src/errors.ts
nitzantomer: I also agree with @doodyparizada that this is a conflict.
The resource might not be gone, it's just that the client is out of sync with the server/db

scripts/src/errors.ts
nitzantomer: i chose `locked` because the resource might be available later on, i.e.: someone opened an order but canceled it before finishing.
all of the other ideas you gave don't convey this.

scripts/src/errors.ts
nitzantomer: this is basically true, BUT i don't want this file (`errors`) to be aware of the `services/orders` file

doodyparizada: I don't care/understand

scripts/src/errors.ts
doodyparizada: If caps are increased? or if the cap is hourly based?
Can 410 change back to 200? i.e. gone and came back?

scripts/src/errors.ts
doodyparizada: We said conflict, because a client will request another order of the same offer only if something got screwed with his state. So that's why we thought conflict conveys that

scripts/src/errors.ts
doodyparizada: Yeah I saw that. So can you not export this class so no one can misuse this?

doodyparizada:
+ travis is failing: ```ERROR: /home/travis/build/kinecosystem/marketplace-server/scripts/src/public/services/orders.ts[119, 34]: Missing semicolon
```

scripts/src/errors.ts
doodyparizada: 404

scripts/src/errors.ts
doodyparizada: algorithm type not supported

nitzantomer:
Closed in favor of: https://github.com/kinecosystem/marketplace-server/pull/89

PR #85, opened 2018-04-29
Doody/return only app offers

doodyparizada:


scripts/src/public/services/offers.ts
```diff
@@ -71,15 +72,19 @@ async function filterOffers(userId: string, offers: db.Offer[], logger: LoggerIn
...
 		offers = offers.concat(
 			await filterOffers(
 				userId,
-				await db.Offer.createQueryBuilder()
-					.where("type = 'earn'")
-					.orderBy("amount", "DESC")
-					.addOrderBy("id", "ASC")
-					.getMany(),
+				(await query
+					.andWhere("offer.type = :type", { type: "earn" })
```
nitzantomer: why not:
```.andWhere("offer.type = 'earn'")```

File: scripts/src/public/services/offers.ts [hunk 2]
```diff
@@ -89,11 +94,11 @@ export async function getOffers(userId: string, appId: string, filters: ModelFil
 		offers = offers.concat(
 			await filterOffers(
 				userId,
-				await db.Offer.createQueryBuilder()
-					.where("type = 'spend'")
-					.orderBy("amount", "ASC")
-					.addOrderBy("id", "ASC")
-					.getMany(),
+				(await query
+					.andWhere("offer.type = :type", { type: "spend" })
```
nitzantomer: same:
```
.andWhere("offer.type = 'spend'")
```

scripts/src/models/applications.ts
```diff
@@ -19,6 +20,10 @@ export class Application extends CreationDateModel {
 
 	@Column("simple-json", { name: "jwt_public_keys" })
 	public jwtPublicKeys!: StringMap;
+
+	@ManyToMany(type => Offer)
```
liorama: Don't you need to specify the owner field on the Offer class?

tests/src/helpers.ts
```diff
@@ -20,11 +21,11 @@ const animalPoll: Poll = {
 	}],
 };
 
-export async function createUser(): Promise<User> {
+export async function createUser(appId?: string): Promise<User> {
 	const uniqueId = generateId();
 	const user = await (User.new({
 		appUserId: `test_${uniqueId}`,
-		appId: "kik",
+		appId: appId || (await Application.findOne())!.id,
```
nitzantomer: what, a random application?

liorama: I'm with Nitzi. I'm also not sure the `await` is needed there.

scripts/src/public/services/offers.ts
doodyparizada: 1. It's safer as you don't have to remember the quotes.
2. It's good practice to use parameters for both performance and security reasons

scripts/src/public/services/offers.ts
doodyparizada: same

tests/src/helpers.ts
doodyparizada: The await is needed. I just want a user attached to some application so it's a valid user. not all tests care about which application the user belongs to, as long as it belongs to one.

scripts/src/create.ts
```diff
@@ -27,17 +27,11 @@ async function createApp(appId: string, apiKey: string, name: string) {
...
 
-	for (const offer of offers) {
-		for (const app of [app2]) {
-			const appOffer = AppOffer.new({
-				appId: app.id,
-				offerId: offer.id
-			});
-			await appOffer.save();
-		}
+	for (const app of [app2]) {
```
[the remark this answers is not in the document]
doodyparizada: I kept adding `kik` and removing it from here. This create script and the create_from_csv should be merged and used as a db init script for the docker system tests (when thats a thing)

scripts/src/models/applications.ts
doodyparizada: not owner, but applications. It's a many-to-many relation between apps (digital services) and offers.
The owner_id is a one-to-many relation and stored in `OfferOwner`.
You don't have to declare the opposite relation if you don't use it.

PR #86, opened 2018-04-29
now 404 is supported in the internal server

nitzantomer:


doodyparizada:
@nitzantomer  I don't understand what was wrong?

PR #88, opened 2018-04-30
save the poll answers

nitzantomer:


scripts/src/utils.ts
```diff
@@ -4,6 +4,11 @@ const fromProjectRoot = _path.join.bind(path, __dirname, "../../");
 
 export type ServerError = Error & { syscall: string; code: string; };
 
+export type Nothing = null | undefined;
+export function isNothing(obj: any): obj is Nothing {
+	return obj === null || obj === undefined;
```
doodyparizada: Is this a standard pattern?

scripts/src/models/offers.ts
```diff
@@ -129,3 +129,19 @@ export class Asset extends CreationDateModel {
 		return Object.assign({ type: this.type }, this.value);
 	}
 }
+
+@Entity({ name: "poll_answers" })
+@Register
+export class PollAnswers extends CreationDateModel {
```
doodyparizada: Maybe keep the class name is single form? `PollAnswer`

scripts/src/utils.ts
nitzantomer: we can just do:
```
return obj == null;
```
but then someone might change it to `===` and that will ruin it.
this way is more explicit, which i prefer

PR #89, opened 2018-04-30
error handling

nitzantomer:
all errors are grouped into one module, and the error middleware knows how to serialize them

### DIFF
--- a/scripts/src/errors.ts
+++ b/scripts/src/errors.ts
@@ -11,7 +11,9 @@ export type ApiError = {
 const CODES = {
 	Unauthorized: {
 		MissingToken: 1,
-		InvalidToken: 2
+		InvalidToken: 2,
+		InvalidApiKey: 3,
+		TOSMissingOrOldToken: 4
 	},
 	NotFound: {
 		App: 1,
@@ -19,6 +21,9 @@ const CODES = {
 		Order: 3,
 		PublicKey: 4,
 		OfferCapReached: 5
+	},
+	RequestTimeout: {
+		OpenOrderExpired: 1
 	},
 	Conflict: {
 		ExternalOrderExhausted: 1
@@ -29,7 +34,8 @@ const CODES = {
 	},
 	BadRequest: {
 		UnknownSignInType: 1,
-		WrongJWTAlgorithm: 2
+		WrongJWTAlgorithm: 2,
+		InvalidPollAnswers: 3
 	}
 };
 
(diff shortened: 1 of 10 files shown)


doodyparizada:
you have an error on travis: ```
scripts/src/mock_client.ts(9,10): error TS2305: Module '"/home/travis/build/kinecosystem/marketplace-server/scripts/src/public/middleware"' has no exported member 'ApiError'.
```

**Comment by nitzantomer** [comment]
@doodyparizada yeah yeah

File: scripts/src/public/services/orders.ts [hunk 1]
```diff
@@ -215,7 +224,7 @@ export async function getOrderHistory(
 
 function openOrderDbToApi(order: db.Order): OpenOrder {
 	if (order.status !== "opened") {
-		throw new Error("only opened orders should be returned");
+		throw OpenedOrdersOnly();
```
doodyparizada: Maybe this should be some assert? it's an internal error and a programming bug

scripts/src/public/services/orders.ts
```diff
@@ -231,7 +240,7 @@ function openOrderDbToApi(order: db.Order): OpenOrder {
 
 function orderDbToApi(order: db.Order): Order {
 	if (order.status === "opened") {
-		throw new Error("opened orders should not be returned");
+		throw OpenedOrdersUnreturnable();
```
doodyparizada: same here: Maybe this should be some assert? it's an internal error and a programming bug

scripts/src/errors.ts
```diff
@@ -0,0 +1,157 @@
+export type ApiError = {
+	code: number;
+	error: string;
+	message: string;
+};
+
+/**
+ * Code additions (/postfix) to be added to the http status code per error.
```
liorama: Why the `/` before postfix? (also, although postfix can be used I think the more correct term is suffix...but I'm just being an Ory now...leave it or change it IDC)

scripts/src/errors.ts
```diff
@@ -0,0 +1,157 @@
...
+function ConflictError(index: number, message: string) {
+	return new MarketplaceError(409, index, "Conflict", message);
+}
+
+export function ExternalOrderExhausted() {
+	return ConflictError(CODES.Conflict.ExternalOrderExhausted, "User already completed offer, or has a pending order");
+}
+
+function LockedError(index: number, message: string) {
+	return new MarketplaceError(423, index, "Resource Locked", message);
```
liorama: Didn't we agree on a 404?

doodyparizada: yes

PR #91, opened 2018-04-30
setting db fields to 'null' instead of 'undefined'

nitzantomer:
setting the fields to `undefined` won't update the db to `null`.
see: https://github.com/typeorm/typeorm/issues/584


doodyparizada:
@nitzantomer this somehow fails the tests

PR #92, opened 2018-05-01
look for key id also in payload

nitzantomer:
until now we looked for it only in the JWT headers, but for Elaz it's easier to add to the payload.
also, throw an error in case `kid` is missing.


liorama:
Elaz says that it might not be a problem and he needs to deal with the error above before he knows if it is or not. So let’s not merge that hack PR before we know it’s absolutely necessary.

nitzantomer:
closed as it was decided not to support `kid` in payload.
but, part of the PR has moved to a new one: https://github.com/kinecosystem/marketplace-server/pull/93

PR #97, opened 2018-05-07
Doody/docker

doodyparizada:
create docker env for orchestrating the whole environment and running the system tests locally.

### Walkthrough:
* To inject parameters into the server I made changes to `config.ts` - environment variables have preference over config file.
* the `Dockerfile` builds the code locally and doesn't declare volumes. The `docker-compose.yaml`, declares a code volume that replaces the pre-built code, to debug during development.
* when running system tests using docker, the current data is dropped from the attached postgres and is re-created from the data in `data/{1,2,3}.csv`. The address and the amount that appear in the data is replaced with values defined in the `tests.yaml` file:
  * `STELLAR_ADDRESS` - replaces the address in offers with this one.
  * `REDUCE_AMOUNT` - reduces the amount in the offers by dividing by 1000 (not to drain the funding account)
* the setup for payment-service is defined in a `./secrets/.secrets` file which docker-compose reads. It includes the private and public key for the funding account.
* because the code is mounted locally, you have to run `make build` for new code to take effect.


.gitignore
```diff
@@ -1,8 +1,9 @@
 .idea
 scripts/bin
-tests/src/**/*.js
+**/*.js
```
nitzantomer: this is risky.
we might need to use plain javascript files someday

scripts/src/middleware.ts
```diff
@@ -116,3 +117,16 @@ function serverErrorHandler(err: any, req: Request, res: Response) {
 	logger.error(message);
 	res.status(500).send({ code: 500, error: err.message || "Server error", message:  err.message });
 }
+
+export const statusHandler = async function(req: express.Request, res: express.Response) {
+	res.status(200).send(
+		{
+			status: "ok",
+			app_name: process.env.APP_NAME,
```
nitzantomer: put this in the config.
everything that comes form the `process.env` should go into the config

scripts/src/middleware.ts
```diff
@@ -116,3 +117,16 @@ function serverErrorHandler(err: any, req: Request, res: Response) {
...
+			start_time: START_TIME,
+			build: {
+				commit: process.env.BUILD_COMMIT,
```
nitzantomer: same

scripts/src/middleware.ts
```diff
@@ -116,3 +117,16 @@ function serverErrorHandler(err: any, req: Request, res: Response) {
...
+				timestamp: process.env.BUILD_TIMESTAMP,
```
nitzantomer: same

.gitignore
doodyparizada: ok. I don't remember why i did this

PR #98, opened 2018-05-10
handle shutdown

nitzantomer:


scripts/src/internal/index.ts
```diff
@@ -15,6 +15,14 @@ server.listen(config.port);
 server.on("error", onError);
 server.on("listening", onListening);
 
+function cleanup() {
+	logger.info("Shutting down");
+	server.close(() => {
+		logger.info("Done, have a great day!");
+		process.exit(1);
```
doodyparizada: if everything is ok, then `exit(0)`

scripts/src/public/index.ts
```diff
@@ -14,6 +14,14 @@ server.listen(config.port);
 server.on("error", onError);
 server.on("listening", onListening);
 
+function cleanup() {
+	logger.info("Shutting down");
+	server.close(() => {
+		logger.info("Done, have a great day!");
+		process.exit(1);
```
doodyparizada: same

doodyparizada: + can this be shared inside the `middleware.ts` file?

PR #99, opened 2018-05-13
createExternalOrder to handle both earn and spend

nitzantomer:


scripts/src/public/jwt.ts
```diff
@@ -5,25 +5,25 @@ import { isNothing } from "../utils";
 import { Application } from "../models/applications";
 import { NoSuchApp, NoSuchPublicKey, JwtKidMissing, WrongJWTAlgorithm } from "../errors";
 
-export type JWTClaims = {
+export type JWTClaims<T extends string> = {
```
doodyparizada: 1. maybe call it SUB instead of T?
2. shouldn't it be `SUB is string`

scripts/src/public/jwt.ts
```diff
@@ -5,25 +5,25 @@ import { isNothing } from "../utils";
...
 	iss: string; // issuer - the app_id
 	exp: number; // expiration
 	iat: number; // issued at
-	sub: string; // subject
+	sub: T; // subject
 };
 
-export type JWTContent<T> = {
+export type JWTContent<T, S extends string> = {
```
doodyparizada: call S -> SUB

scripts/src/public/services/applications.ts
```diff
@@ -14,24 +14,28 @@ export type SignInContext = {
...
 export type SpendPayload = {
-	offer: SpendPayloadOffer;
+	offer: ExternalOfferPayload;
 };
 
-export async function validateSpendJWT(jwt: string, logger: LoggerInstance): Promise<SpendPayloadOffer> {
-	const decoded = await verifyJWT<SpendPayload>(jwt, logger);
-	return decoded.payload.offer;
+export async function validateExternalOrderJWT(jwt: string, logger: LoggerInstance) {
+	const decoded = await verifyJWT<SpendPayload | EarnPayload, "spend" | "earn">(jwt, logger);
```
doodyparizada: This gives a change for an EarnPayload to arrive with "spend" subject. I prefer to be explicit here with the types.

scripts/src/public/services/orders.ts
```diff
@@ -110,13 +112,18 @@ export async function createMarketplaceOrder(offerId: string, user: User, logger
 }
 
 export async function createExternalOrder(jwt: string, user: User, logger: LoggerInstance): Promise<OpenOrder> {
-	const offer = await validateSpendJWT(jwt, logger);
+	const payload = await validateExternalOrderJWT(jwt, logger);
+
+	if (payload.sub === "earn" && (payload as EarnPayload).user_id !== user.id) {
```
doodyparizada: The user_id in the EarnPayload is an `app_user_id` The Digital Service is never exposed to the internal User.id.
You should check `(payload as EarnPayload).user_id !== user.app_user_id`

scripts/src/public/services/orders.ts
```diff
@@ -110,13 +112,18 @@ export async function createMarketplaceOrder(offerId: string, user: User, logger
...
+		throw ExternalEarnOfferByDifferentUser(user.id, (payload as EarnPayload).user_id);
```
doodyparizada: same: `ExternalEarnOfferByDifferentUser(user.app_user_id, ...`

scripts/src/public/services/orders.ts
```diff
@@ -126,7 +133,7 @@ export async function createExternalOrder(jwt: string, user: User, logger: Logge
 			userId: user.id,
 			offerId: offer.id,
 			amount: offer.amount,
-			type: "spend", // TODO: we currently only support native spend
+			type: payload.sub,
```
doodyparizada: is this the only thing that changes?

scripts/src/public/services/orders.ts
```diff
@@ -110,37 +111,46 @@ export async function createMarketplaceOrder(offerId: string, user: User, logger
...
-				title: offer.title,
-				description: offer.description
+				title: payload.offer.title,
+				description: payload.offer.description
 			},
 			blockchainData: {
-				sender_address: user.walletAddress,
-				recipient_address: offer.wallet_address
+				sender_address: payload.sub === "spend" ? user.walletAddress : tmpBlockchainData.sender_address,
+				recipient_address: payload.sub === "spend" ? tmpBlockchainData.recipient_address : user.walletAddress
```
doodyparizada: maybe unite this piece of code with the createMarkerplaceOrder

PR #113, opened 2018-06-04
throw specific error when jwt fails to verify

nitzantomer:


scripts/src/errors.ts
```diff
@@ -175,6 +176,10 @@ export function WrongJWTAlgorithm(type: string) {
 	return BadRequestError(CODES.BadRequest.UnknownSignInType, `algorithm type ("${ type }") not supported`);
 }
 
+export function InvalidJWT() {
```
doodyparizada: InvalidJWTSignature

PR #122, opened 2018-06-12
change kidmissing error code

doodyparizada:


scripts/src/errors.ts
```diff
@@ -14,34 +14,34 @@ const CODES = {
 		MissingToken: 1,
 		InvalidToken: 2,
 		InvalidApiKey: 3,
-		JwtKidMissing: 4,
-		TOSMissingOrOldToken: 5
+		TOSMissingOrOldToken: 5,
```
OrenZak: shouldn't be 4 ?

scripts/src/errors.ts
```diff
@@ -14,34 +14,34 @@ const CODES = {
...
 	},
 	BadRequest: {
 		UnknownSignInType: 1,
 		WrongJwtAlgorithm: 2,
-		InvalidJwtSignature: 3,
-		InvalidPollAnswers: 4,
-		InvalidExternalOrderJwt: 5
+		InvalidPollAnswers: 3,
+		InvalidExternalOrderJwt: 4,
+		InvalidJwtSignature: 5,
```
OrenZak: `InvalidJwtSignature` is not written on the Wiki

scripts/src/errors.ts
doodyparizada: I didn't want to break the current number. Are you or @Elaz  using this error?

scripts/src/errors.ts
doodyparizada: added

scripts/src/errors.ts
OrenZak: nop, you can change

PR #136, opened 2018-06-25
removed event, added another one

nitzantomer:
added: `earn_transaction_broadcast_to_blockchain_submitted`
removed: `wallet_creation_succeeded`


scripts/src/public/services/orders.ts
```diff
@@ -244,6 +247,7 @@ export async function submitOrder(
 	logger.info("order changed to pending", { orderId });
 
 	if (order.type === "earn") {
+		createEarnTransactionBroadcastToBlockchainSubmitted(order.userId, order.offerId, order.id).report();
```
doodyparizada: change order

PR #141, opened 2018-07-02
added the origin to the order result

nitzantomer:
also, kinda changed the way it's built


scripts/src/public/services/orders.ts
```diff
@@ -317,7 +319,18 @@ function orderDbToApi(order: db.Order): Order {
...
+	const apiOrder = Object.assign(
+		pick(order, "id", "origin", "status", "amount"), {
+			result: order.value,
+			offer_type: order.type,
+			offer_id: order.offerId,
+			title: order.meta.title,
+			error: order.error as ApiError,
+			blockchain_data: order.blockchainData,
+			completion_date: (order.currentStatusDate || order.createdDate).toISOString()
+		}, order.meta) as Order;
```
doodyparizada: be more explicit

PR #142, opened 2018-07-02
hopefully it will fix it, or give us more info to work with

nitzantomer:


scripts/src/middleware.ts
```diff
@@ -75,7 +75,8 @@ export const reportMetrics = function(req: express.Request, res: express.Respons
 	const t = performance.now();
 
 	res.on("finish", () => {
-		metrics.timeRequest(performance.now() - t, req.method, req.route);
+		const path = req.route ? req.route.path : (req.url || "unknown");
```
doodyparizada: We can try this for a short period of time. We don't want to overload datadog with hundreds of tags

PR #147, opened 2018-07-09
Eco 494 disable whitelist signin

doodyparizada:


scripts/src/config.ts
```diff
@@ -4,6 +4,7 @@ import "source-map-support/register";
 import { path } from "./utils";
 
 import { LogTarget } from "./logging";
+import { SignInType } from "./models/applications";
```
nitzantomer: **NO NO NO !!!**

the config must know of no other module that might depend on him later on, we want to avoid circular dependencies.
so, either define `SignInType` here and use it in the  `models/applications` module, or create a new module which holds the types and then both will use that.

config/internal.default.json
```diff
@@ -20,5 +20,6 @@
 	"payment_service": "http://localhost:5000",
 	"internal_service": "http://localhost:3001",
 	"jwt_keys_dir": "jwt",
+	"sign_in_types": ["jwt", "whitelist"],
```
nitzantomer: i'd feel more comfortable if the default is always `jwt` and you'll need to opt-in for it

config/public.default.json
```diff
@@ -29,5 +29,6 @@
 	"ecosystem_service": "http://localhost:3000",
 	"environment_name": "local",
 	"bi_service": "https://kin-bi.appspot.com/eco_",
+	"sign_in_types": ["jwt", "whitelist"],
```
nitzantomer: same, default to `jwt`

data/apps/test.json
```diff
@@ -0,0 +1,13 @@
+{
+  "api_key": "AyINT44OAKagkSav2vzMz",
+  "jwt_public_keys": {
+    "rs512_0": "-----BEGIN PUBLIC KEY-----\nMIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQCgOMeb0heN3k9NinbsWpRo3WEl\nwGnoP9UCpKsGLbZAzkgE64fP/thj4CH6xnoRsObl4LFdeh9cekXooS19/A+Q9M0k\nxlrURq8qvogJPSFERi6S7vzMM1LxGmOcgwLLSp4AGZ5A0GkjFXhiZU1hOfb2WmkK\neVTyx22Df6Ox6ulslwIDAQAB\n-----END PUBLIC KEY-----",
+    "es256_0": "-----BEGIN PUBLIC KEY-----\nMFYwEAYHKoZIzj0CAQYFK4EEAAoDQgAE5ppEUgBzRQpccf9arl/txo8t6hDpGPnK\n1SxVTpadjrlCi4YmKM4xVEzyYSsOQHvWA6I70gT0DyVxSk64zsY0MA==\n-----END PUBLIC KEY-----"
+  },
+  "config": {
+    "max_user_wallets": -1,
+    "sign_in_types": ["jwt", "whitelist"]
```
nitzantomer: here it's fine

scripts/src/create.ts
```diff
@@ -217,17 +215,23 @@ function getStellarAddresses() {
 }
 
 initModels().then(async () => {
-	// create apps
-	const app1 = await createApp("smpl", "Sample Application", ["default", "1"], Application.SAMPLE_API_KEY);
-	const app2 = await createApp("kik", "Kik Messenger", ["1"]);
-	const app3 = await createApp("test", "Test App", ["es256_0", "rs512_0"]);
+
+	const appsDir = process.argv[2];
```
nitzantomer: maybe you want to use something like: https://github.com/tj/commander.js

scripts/src/models/applications.ts
```diff
@@ -1,9 +1,16 @@
 import { generateId, IdPrefix } from "../utils";
 import { Column, Entity, Index, JoinTable, ManyToMany } from "typeorm";
-import { CreationDateModel, Model, register as Register, initializer as Initializer } from "./index";
+import { CreationDateModel, register as Register, initializer as Initializer } from "./index";
 import { Offer } from "./offers";
+import { getConfig } from "../config";
 
+const NO_WALLET_LIMIT = -1;
```
nitzantomer: why `-1` instead of `null`? what is this, c?

tests/src/helpers.ts
```diff
@@ -174,7 +174,8 @@ export async function createApp(appId: string): Promise<Application> {
 		id: appId,
 		name: appId,
 		jwtPublicKeys: {},
-		walletAddresses: { recipient: address, sender: address }
+		walletAddresses: { recipient: address, sender: address },
+		config: { max_user_wallets: -1, sign_in_types: ["jwt", "whitelist"] }
```
nitzantomer: `null`!!

scripts/src/config.ts
doodyparizada: I think I'll just define it as plain string. There is no enforcement here anyway.

config/internal.default.json
doodyparizada: actually sign_in_type should only be defined for public endpoint, but because the shared model uses it, it has to be defined in both. Mabye I'll pass the global sign_in_type config as a parameter to the model function instead

config/public.default.json
doodyparizada: This is just an example. not a default

PR #149, opened 2018-07-29
ECO-522 auto restart server

liorama:


scripts/src/metrics.ts
```diff
@@ -45,6 +45,10 @@ export function reportServerError(method: string, path: string) {
 	statsd.increment("server_error", 1, undefined, { method, path });
 }
 
+export function reportProcessAbort() {
```
doodyparizada: maybe attach a reason instead of system: exit? i.e. reason: "multiple 5xx errors in short period"

doodyparizada: the reason can be a parameter

scripts/src/middleware.ts
```diff
@@ -8,8 +8,14 @@ import { getConfig } from "./config";
 import { generateId, pick } from "./utils";
 import { MarketplaceError } from "./errors";
 import { getDefaultLogger } from "./logging";
+import { abort as restartServer } from "./server";
 
 const START_TIME = (new Date()).toISOString();
+
+const RESTART_ERROR_COUNT = 5;  // Amount io errors to occur in time frame to trigger restart
```
doodyparizada: why io errors? can't it be any unknown error that causes 5xx?

scripts/src/middleware.ts
```diff
@@ -113,9 +119,12 @@ function clientErrorHandler(error: MarketplaceError, req: express.Request, res:
 
 function serverErrorHandler(error: any, req: express.Request, res: express.Response) {
 	const log = req.logger || logger;
-
 	metrics.reportServerError(req.method, req.url);
 
+	const timestamp = Math.floor(Date.now() / 1000);
```
doodyparizada: `moment.utc().seconds()` or  `new Date().getUTCSeconds();`

scripts/src/middleware.ts
```diff
@@ -129,6 +138,12 @@ function serverErrorHandler(error: any, req: express.Request, res: express.Respo
 		message += `message: ${ error.toString() }`;
 	}
 
+	if (serverErrorTimeStamps.length === RESTART_ERROR_COUNT) {
```
doodyparizada: if you initialize `serverErrorTimeStamps` to 20 zeros, you can remove this `if`

docker-compose.dev.yaml
```diff
@@ -5,6 +5,7 @@ services:
   marketplace-server:
     image: kinecosystem/marketplace-server:latest
     command: ./tests/wait-for marketplace-internal:3000 -- npm run start
+    restart: no
```
doodyparizada: `no` translate to `false` in yaml, so you need parenthesis `"no"`. That's why travis is failing

scripts/src/middleware.ts
liorama: lol, spelling mistake. suppose to be "of"

scripts/src/middleware.ts
liorama: Both are not good, they give the seconds value of the current time. So you are limited to a timeframe lower than a minute and if more than a minute passes you got a bug on your hands.
If anything `moment().unix()` fits better.

scripts/src/middleware.ts
liorama: We can minify the code if we just want to make it less readable.

doodyparizada:
looks good, but travis failed. maybe @nitzantomer can take a look?

PR #150, opened 2018-07-31
Eco 549

doodyparizada:
add timing to order completion metric


scripts/src/models/orders.ts
```diff
@@ -203,8 +203,8 @@ export class Order extends CreationDateModel {
 	@Column()
 	public status!: OpenOrderStatus;
 
-	@Column({ name: "current_status_date", nullable: true })
-	public currentStatusDate?: Date;
+	@Column({ name: "current_status_date" })
```
doodyparizada: This has a default now, so it isn't null

PR #152, opened 2018-08-01
Nitzan/use kin.js client

nitzantomer:


package.json
```diff
@@ -8,14 +8,14 @@
     "url": "https://github.com/kinfoundation/marketplace-server"
   },
   "dependencies": {
+    "@kinecosystem/kin.js": "^1.0.3",
```
doodyparizada: why do we have a `@kinecosystem` prefix?

scripts/src/client.ts
```diff
@@ -0,0 +1,239 @@
+import * as uuid4 from "uuid4";
+// import { Keypair } from "stellar-sdk";
```
doodyparizada: delete

scripts/src/client.ts
```diff
@@ -0,0 +1,239 @@
...
+import { KinWallet, createWallet, KinNetwork, Payment, Keypair } from "@kinecosystem/kin.js";
```
doodyparizada: why do you not just import `"kin"`?

scripts/src/client.ts
```diff
@@ -0,0 +1,239 @@
...
+	}
+}
+
+export class ClientError extends Error {
+	public response?: AxiosResponse;
+}
+
+class ClientRequests {
+	public static async create(data: { device_id: string; wallet_address: string; }) {
+			const res = await axios.post<AuthToken>(MARKETPLACE_BASE + "/v1/users", data);
```
doodyparizada: seems like double spacing

PR #156, opened 2018-08-05
refactored orders to work with a cross table

nitzantomer:
p2p functionality is mostly done and current unit tests are passing


scripts/src/models/orders.ts
```diff
@@ -241,6 +305,98 @@ export class Order extends CreationDateModel {
...
+	@ManyToOne(type => Order, order => order.contexts)
+	@JoinColumn({ name: "order_id" })
+	public readonly order!: Order;
+
+	@ManyToOne(type => User, user => user.contexts)
+	@JoinColumn({ name: "user_id" })
+	public readonly user!: User;
+
+	@Column()
+	public readonly role!: "sender" | "recipient";
```
doodyparizada: type => earn/ spend

scripts/src/public/routes/orders.ts
```diff
@@ -33,6 +33,7 @@ export type CreateExternalOrderRequest = Request & {
  * create an order for a native offer
  */
 export const createExternalOrder = async function(req: CreateExternalOrderRequest, res: Response) {
+	console.log("createExternalOrder");
```
doodyparizada: delete this

scripts/src/public/services/orders.ts
```diff
@@ -302,37 +326,50 @@ export async function getOrderHistory(
 	};
 }
 
-function openOrderDbToApi(order: db.Order): OpenOrder {
+function getMeta(order: db.NormalOrder | db.P2POrder, userId: string) {
```
doodyparizada: you can just access the contexts of the relevant userId instead of using the helper isXXX functions

scripts/src/public/services/orders.ts
```diff
@@ -341,8 +378,15 @@ export async function setFailedOrder(order: db.Order, error: MarketplaceError, f
 	order.setStatus("failed");
 	order.currentStatusDate = failureDate || order.currentStatusDate;
 	order.error = error.toJson();
-	const user = await User.findOneById(order.userId);
-	metrics.orderFailed(order, user);
+
+	if (order.origin === "p2p") {
+		// TODO: which user to send? maybe both?
```
doodyparizada: actually both is a good idea. because sender might fail due to timeouts, and recipient might fail because account wasn't created for example. This is just for the error message that will be printed into datadog

tests/src/helpers.ts
```diff
@@ -39,69 +39,77 @@ export async function createUser(appId?: string): Promise<User> {
...
 		}
 	});
+	await order.save();
+
+	const context = OrderContext.new({
+		userId,
+		user,
+		order,
+		orderId: order.id,
+		role: "recipient",
```
doodyparizada: should be same as offer.type

tests/src/helpers.ts
```diff
@@ -110,6 +118,20 @@ export async function createExternalOrders(userId: string): Promise<number> {
...
+		meta: {
+			title: "external order #1",
+			description: "first external order"
+		}
+	});
+	await context.save();
+	order.contexts.push(context);
```
doodyparizada: DRY on creation of Order and context

scripts/src/admin/services.ts
```diff
@@ -224,22 +224,28 @@ async function offerToHtml(offer: Offer): Promise<string> {
...
+	let html = "";
+
+	for (const context of order.contexts) {
+		html += `<tr>
+<td><a href="/orders/$ {order.id }">${ order.id }</a></td>
+<td class="status_${ order.status }"><a href="/orders?status=${ order.status }">${ order.status }</a></td>
+<td><pre>${ JSON.stringify(order.error) }</pre></td>
+<td>${ order.origin }</td>
+<td>${ order.type }</td>
+<td><a href="/users/${ context.user.id }">${ context.user.id }</a></td>
```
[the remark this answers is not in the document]
nitzantomer: it's as [it is now](https://github.com/kinecosystem/marketplace-server/blob/master/scripts/src/admin/services.ts#L233):
```
<td><a href="/users/${order.userId}">${order.userId}</a></td>
```

scripts/src/models/orders.ts
```diff
@@ -70,51 +96,88 @@ export class Order extends CreationDateModel {
...
-			.andWhere("user_id = :userId", { userId })
-			.andWhere("expiration_date > :date", { date: latestExpiration })
-			.orderBy("expiration_date", "DESC"); // if there are a few, get the one with the most time left
+
+		const query = Order.createQueryBuilder("ordr") // don't use 'order', it messed things up
+			.leftJoinAndSelect("ordr.contexts", "context")
+			.leftJoinAndSelect("context.user", "user")
+			.where("ordr.offer_id = :offerId", { offerId })
+			.andWhere("ordr.status = :status", { status: "opened" })
+			.andWhere("context.user_id = :userId", { userId })
```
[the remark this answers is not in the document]
nitzantomer: where am i using `contenxt.id`?

scripts/src/models/orders.ts
```diff
@@ -70,51 +96,88 @@ export class Order extends CreationDateModel {
...
+			.andWhere("ordr.expiration_date > :date", { date: latestExpiration })
+			.orderBy("ordr.expiration_date", "DESC"); // if there are a few, get the one with the most time left
+
+		const order = await (query.getOne() as Promise<T | undefined>);
+
+		if (order && !order.contexts) {
+			order.contexts = [];
```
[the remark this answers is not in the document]
nitzantomer: why?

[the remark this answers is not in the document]
doodyparizada: Isn't this an unexpected case?

scripts/src/admin/services.ts
```diff
@@ -428,22 +428,16 @@ export async function getOrders(params: any, query: Paging & { status?: OpenOrde
...
 	}
 
-	// TODO: what is this for?
-	/*const user = await User.findOneById(order.userId);
-	if (!user) {
-		throw new Error("user not found: " + order.userId);
-	}*/
-
-	await payment.payTo(order.blockchainData.recipient_address!, order.sender!.appId, order.amount, order.id, getDefaultLogger());
+	await payment.payTo(order.blockchainData.recipient_address!, order.contexts[0].user.appId, order.amount, order.id, getDefaultLogger());
```
doodyparizada: why not contextFor?

scripts/src/internal/services.ts
```diff
@@ -91,11 +92,13 @@ export async function paymentComplete(payment: CompletedPayment, logger: LoggerI
...
 	}
 
-	if (order.type === "earn") {
-		createEarnTransactionBroadcastToBlockchainSucceeded(order.recipient!.appUserId, payment.transaction_id, order.offerId, order.id).report();
-	} else {
-		createSpendOrderPaymentConfirmed(order.sender!.appUserId, payment.transaction_id, order.offerId, order.id, order.isExternalOrder(), order.origin).report();
-	}
+	order.forEachContext(context => {
+		if (context.type === "earn") {
+			createEarnTransactionBroadcastToBlockchainSucceeded(context.user.id, payment.transaction_id, order.offerId, order.id).report();
```
doodyparizada: Ask Oded if you need this for p2p

scripts/src/internal/services.ts
```diff
@@ -132,11 +135,10 @@ export async function paymentComplete(payment: CompletedPayment, logger: LoggerI
 	if (!payment.app_id) {
 		logger.error(`payment is missing the app_id <${ payment.id }, ${ payment.transaction_id }> - setting the one from the DB`);
 
-		const userId = order.type === "earn" ? order.recipient!.id : order.sender!.id;
-		const user: User | undefined = await User.findOneById(userId);
+		const user = order.contexts[0].user;
 
 		if (!user) {
-			logger.error(`failed to fix missing app_id on payment - cant find user ${ userId }`);
+			logger.error(`failed to fix missing app_id on payment - cant find user ${ order.contexts[0].user.id }`);
```
doodyparizada: contextFor?

scripts/src/internal/services.ts
```diff
@@ -132,11 +135,10 @@ export async function paymentComplete(payment: CompletedPayment, logger: LoggerI
 	if (!payment.app_id) {
 		logger.error(`payment is missing the app_id <${ payment.id }, ${ payment.transaction_id }> - setting the one from the DB`);
 
-		const userId = order.type === "earn" ? order.recipient!.id : order.sender!.id;
-		const user: User | undefined = await User.findOneById(userId);
+		const user = order.contexts[0].user;
```
doodyparizada: contextFor

scripts/src/internal/services.ts
```diff
@@ -154,14 +156,14 @@ export async function paymentComplete(payment: CompletedPayment, logger: LoggerI
 				return;
 			} else {
 				order.value = asset.asOrderValue();
-				asset.ownerId = order.sender!.id;
+				asset.ownerId = order.user.id;
```
doodyparizada: don't you have sender anymore?

scripts/src/internal/services.ts
```diff
@@ -154,14 +156,14 @@ export async function paymentComplete(payment: CompletedPayment, logger: LoggerI
...
 			}
 		}
-	} else if (order.isExternalOrder()) {
-		// XXX for p2p don't put the JWT in the recipient order's value
-		// XXX for p2p create a completed order for the recipient too
-		order.value = await getPaymentJWT(order, payment.app_id, order.type === "spend" ? order.sender!.id : order.recipient!.id);
+	} else if (order.isP2P()) {
+		order.value = await getPaymentJWT(order, payment.app_id, order.sender.id);
+	} else if (order.isNormal()) {
+		order.value = await getPaymentJWT(order, payment.app_id, order.user.id);
```
doodyparizada: Aren't these two if statements doing the same thing?

doodyparizada: Not sure this is the correct behaviour

scripts/src/internal/services.ts
```diff
@@ -175,12 +177,15 @@ export async function paymentComplete(payment: CompletedPayment, logger: LoggerI
 	order.setStatus("completed");
 	await order.save();
 
-	metrics.completeOrder(order.type, order.offerId, prevStatus, (order.currentStatusDate.getTime() - prevStatusDate.getTime()) / 1000);
+	order.forEachContext(context => {
+		metrics.completeOrder(context.type, order.offerId, prevStatus, (order.currentStatusDate.getTime() - prevStatusDate.getTime()) / 1000);
```
doodyparizada: Ask Oded if this needs to be reported twice

scripts/src/metrics.ts
```diff
@@ -64,23 +64,24 @@ export function orderFailed(order: Order, relatedUser?: User) {
...
-	const title = safeString(error.message);
-	statsd.event(title, message,
-		{ alert_type: "warning" },
-		{
-			order_type: order.type,
-			app_id: user.appId,
-			order_id: order.id,
-			order_origin: order.origin,
-			type: "failed_order" });
+		const title = safeString(error.message);
```
doodyparizada: No need to submit two events. Just one with this line doubled:
```
UserId: ${ context.user.id } | AppId: <${ context.user.appId }> | UserAppId: ${ context.user.appUserId }
```

scripts/src/public/services/orders.ts
```diff
@@ -151,65 +143,100 @@ export async function createMarketplaceOrder(offerId: string, user: User, logger
...
-			user, order,
-			orderId: order.id,
-			userId: user.id,
-			meta: {
-				title,
-				description
-			},
-			role: payload.sub === "spend" ? "sender" : "recipient"
+		order.contexts.forEach(context => {
+			metrics.createOrder("external", context.type, payload.offer.id);
```
doodyparizada: I'd only report this once and change the type to p2p here

scripts/src/public/services/orders.ts
```diff
@@ -250,14 +277,15 @@ export async function submitOrder(
...
+
 			switch (offerContent.contentType) {
 				// TODO this switch-case should be inside the offerContents module
 				case "poll":
 					// validate form
 					if (!offerContents.isValid(offerContent, form)) {
 						throw InvalidPollAnswers();
 					}
-					await offerContents.savePollAnswers(order.recipient!.id, order.offerId, orderId, form); // TODO should we also save quiz results?
+					await offerContents.savePollAnswers(order.user.id, order.offerId, orderId, form); // TODO should we also save quiz results?
```
doodyparizada: why not recipient?

scripts/src/public/services/orders.ts
```diff
@@ -276,12 +304,15 @@ export async function submitOrder(
 	await order.save();
 	logger.info("order changed to pending", { orderId });
 
-	if (order.type === "earn") {
```
doodyparizada: Just because of this check, I think that the order should contain the nature of the flow and not infer it.

Either add the origin/order_type into order or down here check that there is only a single context and that it is earn (instead of checking for existence of earn)

scripts/src/public/services/orders.ts
```diff
@@ -276,12 +304,15 @@ export async function submitOrder(
...
+	const earnContext = order.contexts.find(context => context.type === "earn");
+	if (earnContext) {
 		await payment.payTo(walletAddress, appId, order.amount, order.id, logger);
-		createEarnTransactionBroadcastToBlockchainSubmitted(order.recipient!.id, order.offerId, order.id).report();
+		createEarnTransactionBroadcastToBlockchainSubmitted(earnContext.user!.id, order.offerId, order.id).report();
```
doodyparizada: why not recipient

scripts/src/public/services/orders.ts
```diff
@@ -276,12 +304,15 @@ export async function submitOrder(
...
 	}
 
-	metrics.submitOrder(order.type, order.offerId);
+	order.contexts.forEach(context => {
+		metrics.submitOrder(context.type, order.offerId);
```
doodyparizada: send once like above with the order type (earn,spend,p2p), maybe call it `flow`

scripts/src/public/services/orders.ts
```diff
@@ -341,11 +372,13 @@ function openOrderDbToApi(order: db.Order, userId: string): OpenOrder {
 		throw OpenedOrdersOnly();
 	}
 
+	const context = order.contextFor(userId)!;
 	const meta = getMeta(order as db.P2POrder | db.NormalOrder, userId);
```
doodyparizada: Why isn't getMeta a method of Order?

scripts/src/public/services/orders.ts
```diff
@@ -379,13 +413,9 @@ export async function setFailedOrder(order: db.Order, error: MarketplaceError, f
...
 
-	if (order.origin === "p2p") {
-		// TODO: which user to send? maybe both?
-	} else if (order.type === "earn") {
-		metrics.orderFailed(order, await User.findOneById(order.recipient!.id));
-	} else {
-		metrics.orderFailed(order, await User.findOneById(order.sender!.id));
-	}
+	order.contexts.forEach(context => {
+		metrics.orderFailed(order);
```
doodyparizada: once

scripts/src/public/services/orders.ts
```diff
@@ -359,12 +392,13 @@ function orderDbToApi(order: db.Order, userId: string): Order {
 		throw OpenedOrdersUnreturnable();
 	}
 
+	const context = order.contextFor(userId)!;
 	const meta = getMeta(order as db.P2POrder | db.NormalOrder, userId);
```
doodyparizada: same:
> Why isn't getMeta a method of Order?

scripts/src/admin/services.ts
nitzantomer: because you need the user id for that, which we don't have here

scripts/src/internal/services.ts
nitzantomer: again, the function needs the user id

scripts/src/internal/services.ts
nitzantomer: no user id, no contextFor

scripts/src/internal/services.ts
nitzantomer: there's a `sender`, but when there's one user you can just user that

scripts/src/internal/services.ts
nitzantomer: almost the same, but one uses `order.sender.id` while the other uses `order.user.id`.
when using `order.user` you don't care about whether he's the sender or receiver.

scripts/src/metrics.ts
nitzantomer: what about the `order_type`

scripts/src/public/services/orders.ts
nitzantomer: what about the type?

scripts/src/public/services/orders.ts
nitzantomer: the `order.isMarketplaceOrder()` type guard, when returning true, indicates that order is of type `MarketplaceOrder`.
that type always has only one user, so it exposes that `user` property.
there's currently no type guard for `isEarn()` which will provide a `recipient` property (it does exists for `isP2P()`)

scripts/src/public/services/orders.ts
nitzantomer: the `OrderContext` object has no `recipient`, only `user`

scripts/src/internal/services.ts
```diff
@@ -63,7 +63,8 @@ export type JWTBodyPaymentConfirmation = {
 };
 
 async function getPaymentJWT(order: db.Order, appId: string, userId: string): Promise<OrderValue> {
-	const user = (await User.findOneById(userId))!;
+	// const user = (await User.findOneById(userId))!;
```
doodyparizada: delete

scripts/src/internal/services.ts
```diff
@@ -72,12 +73,14 @@ async function getPaymentJWT(order: db.Order, appId: string, userId: string): Pr
...
+	if (order.isP2P()) {
+		payload.sender_user_id = order.sender.appUserId;
+		payload.recipient_user_id = order.recipient.appUserId;
+	} else if (loggedInContext.type === "earn") {
+		payload.recipient_user_id = loggedInContext.user.appUserId;
 	} else {
-		payload.sender_user_id = user.appUserId;
+		payload.sender_user_id = loggedInContext.user.appUserId;
 	}
-	// XXX if it's p2p, add both recipient and sender user_ids
```
doodyparizada: deleted XXX. Yay!

scripts/src/mock_client.ts
```diff
@@ -42,6 +42,20 @@ class SampleAppClient {
...
+	public async getP2PJWT(data: {
+		offer_id: string;
+		amount: number;
+		sender_title: string;
+		sender_description: string;
+		recipient_id: string;
+		recipient_title: string;
+		recipient_description: string; }) {
+
+		const datastr = Object.keys(data).map(key => `${ key }=${ data[key as keyof typeof data] }`).join("&");
```
doodyparizada:
```
const datastr = Object.keys(data).map(key: keyof typeof data => `${ key }=${ data[key] }`).join("&");
```

scripts/src/mock_client.ts
```diff
@@ -428,8 +441,63 @@ async function nativeEarnFlow() {
 }
 
 async function p2p() {
-	const senderId = "test:rich_user:" + generateId();
+	const offer = {
+		id: "ofer-id",
```
doodyparizada: offer

scripts/src/mock_client.ts
```diff
@@ -443,6 +511,7 @@ async function main() {
 	await didNotApproveTOS();
 	await testRegisterNewUser();
 	await tryToNativeSpendTwice();
+	p2p();
```
doodyparizada: await

scripts/src/models/orders.ts
```diff
@@ -47,15 +47,18 @@ function updateQueryWithStatus(query: SelectQueryBuilder<any>, status?: OrderSta
...
 }
 
-function updateQueryWithFilter(query: SelectQueryBuilder<any>, name: string, value?: string | null) {
+function updateQueryWithFilter(query: SelectQueryBuilder<any>, name: string, value?: string | null, alias?: string) {
 	if (!value) {
 		return;
 	}
 
+	// in case the query is using table alias names, use it with status
+	const fieldName = alias ? `${ alias }.${ name }` : name;
```
doodyparizada: when is passing the alias a must? and when is it optional?

scripts/src/models/orders.ts
```diff
@@ -70,51 +96,88 @@ export class Order extends CreationDateModel {
...
 		const statuses = userId ? ["pending"] : ["opened", "pending"];
 
-		const query = Order.createQueryBuilder()
-			.where("offer_id = :offerId", { offerId })
-			.andWhere(new Brackets(qb => {
-				qb.where("status = :status", { status: "completed" })
+		const query = Order.createQueryBuilder("ordr"); // don't use 'order', it messed things up
+		if (userId) {
+			query
+				.innerJoin("ordr.contexts", "context")
```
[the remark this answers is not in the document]
doodyparizada: @nitzantomer did you check this?

scripts/src/models/orders.ts
```diff
@@ -337,7 +339,9 @@ class OrderImpl extends CreationDateModel implements Order {
 	@Column("simple-json", { name: "blockchain_data", nullable: true })
 	public blockchainData!: BlockchainData;
 
-	@OneToMany(type => OrderContext, context => context.order)
+	@OneToMany(type => OrderContext, context => context.order, {
+		cascadeInsert: true
```
doodyparizada: https://github.com/typeorm/typeorm/blob/065e0ceb136d51533e57c0324d271f0bbada70eb/CHANGELOG.md#020:
> replaced cascadeInsert with cascade: ["insert"] syntax from relation options.

scripts/src/models/orders.ts
```diff
@@ -486,6 +484,21 @@ class OrderImpl extends CreationDateModel implements Order {
 
 		return this.contexts[0].type;
 	}
+
+	@Transaction()
```
doodyparizada: This looks crazy

scripts/src/public/services/orders.ts
```diff
@@ -108,7 +109,7 @@ async function createOrder(offer: offerDb.Offer, user: User) {
 		}
 	}, {
 		user,
-		userId: user.id,
+		// userId: user.id,
```
doodyparizada: delete

scripts/src/public/services/orders.ts
```diff
@@ -144,12 +145,13 @@ export async function createMarketplaceOrder(offerId: string, user: User, logger
...
 async function createP2PExternalOrder(sender: User, jwt: ExternalPayToUserOrderJwt): Promise<db.ExternalOrder> {
-	const recipient = await User.findOneById(jwt.recipient.user_id);
+	const recipient = await User.findOne({ appId: sender.appId, appUserId: jwt.recipient.user_id });
 
 	if (!recipient) {
-		throw new Error(`couldn't find user with id "${ jwt.recipient.user_id }"`);
+		throw NoSuchUser(jwt.recipient.user_id);
 	}
 
+	await addWatcherEndpoint([recipient.walletAddress]);
```
doodyparizada: Great, we should add what Ron started to work on with dynamic watchers on payment service

scripts/src/public/services/orders.ts
```diff
@@ -161,12 +163,12 @@ async function createP2PExternalOrder(sender: User, jwt: ExternalPayToUserOrderJ
 	}, {
 		type: "earn",
 		user: recipient,
-		userId: recipient.id,
+		// userId: recipient.id,
```
doodyparizada: delete

scripts/src/public/services/orders.ts
```diff
@@ -161,12 +163,12 @@ async function createP2PExternalOrder(sender: User, jwt: ExternalPayToUserOrderJ
...
 		meta: pick(jwt.recipient, "title", "description")
 	}, {
 		type: "spend",
 		user: sender,
-		userId: sender.id,
+		// userId: sender.id,
```
doodyparizada: delete

scripts/src/public/services/orders.ts
```diff
@@ -189,7 +191,7 @@ async function createNormalEarnExternalOrder(recipient: User, jwt: ExternalEarnO
 	}, {
 		type: "earn",
 		user: recipient,
-		userId: recipient.id,
+		// userId: recipient.id,
```
doodyparizada: delete

scripts/src/public/services/orders.ts
```diff
@@ -212,7 +216,7 @@ async function createNormalSpendExternalOrder(sender: User, jwt: ExternalSpendOr
 	}, {
 		type: "spend",
 		user: sender,
-		userId: sender.id,
+		// userId: sender.id,
```
doodyparizada: delete

scripts/src/mock_client.ts
nitzantomer: yeah, that doesn't work, i tried it but the compiler doesn't like it (the keys are string as far as it is concerned, and there's no way to convince it to narrow it down

scripts/src/mock_client.ts
nitzantomer: oh right, that explains it

scripts/src/models/orders.ts
nitzantomer: but that's for typeorm version > `2.0`, and we're (currently) using `0.1.12`

scripts/src/models/orders.ts
```diff
@@ -234,34 +393,125 @@ export class Order extends CreationDateModel {
...
+
+		return this.contexts[0].type;
+	}
+}
+
+@Entity({ name: "orders_contexts" })
+@Register
+export class OrderContext extends BaseEntity {
+	public static new(this: ObjectType<OrderContext>, data?: DeepPartial<OrderContext>): OrderContext {
+		return (this as typeof BaseEntity).create(data!) as OrderContext;
```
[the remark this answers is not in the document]
nitzantomer: I tried it at first using `create` but that did not end well, but I don't remember exactly what the problem was.  `new` it is

PR #160, opened 2018-08-16
remove missing app_id hack

doodyparizada:


nitzantomer:
i don't understand this PR.
are you saying that this scenario is no longer possible, or that it never was?

doodyparizada:
It was a quick fix for a bug in iOS which Elaz said was fixed in a newer version.
I left it in for a while to see that it really doesn't happen

nitzantomer:
ok, next time when you open a PR for something that unclear, please be sure to put that explanation in the PR description. merci

PR #162, opened 2018-08-27
temp code to bridge between the old and new implementation of Order

nitzantomer:
# DO NOT MERGE

This PR is just for the review, but it should not be merged, instead we'll merge https://github.com/kinecosystem/marketplace-server/pull/156

### DIFF
--- a/scripts/src/models/orders.ts
+++ b/scripts/src/models/orders.ts
@@ -1,10 +1,21 @@
 import * as moment from "moment";
 import { DeepPartial } from "typeorm/common/DeepPartial";
-import { BaseEntity, Brackets, Column, Entity, getManager, Index, SelectQueryBuilder } from "typeorm";
-import { generateId, IdPrefix } from "../utils";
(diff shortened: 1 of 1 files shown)


scripts/src/models/orders.ts
```diff
@@ -259,6 +270,25 @@ export class Order extends CreationDateModel {
...
+
+	public save() {
+		const content = OrderContext.create({
+			type: this.type,
+			meta: this.meta,
+			orderId: this.id,
+			userId: this.userId
+		});
+
+		return Promise.all([super.save(), content.save()]).then(_ => this);
```
doodyparizada: transaction

doodyparizada: I like `() =>`

scripts/src/models/orders.ts
```diff
@@ -259,6 +270,25 @@ export class Order extends CreationDateModel {
...
+	}
+
+	public remove() {
```
doodyparizada: transaction

PR #164, opened 2018-08-29
delete contexts before deleteing order

nitzantomer:


scripts/src/models/orders.ts
```diff
@@ -530,6 +530,13 @@ class OrderImpl extends CreationDateModel implements Order {
 
 		return this.contexts[0].type;
 	}
+
+	public remove() {
+		return getManager().transaction(async manager => {
+			await Promise.all((await OrderContext.find({ order: this })).map(context => manager.remove(context)));
```
doodyparizada: @nitzantomer shouldn't it be find({ orderId: this.orderId })?

doodyparizada: maybe await OrderContext.find({ order: this }) found all orders in the DB?

PR #165, opened 2018-08-29
Temp fix admin

doodyparizada:


scripts/src/admin/services.ts
```diff
@@ -405,22 +411,23 @@ export async function getApplicationUserData(params: { app_user_id: string, app_
...
 	}
-	const orders = await Order.find({
-		where: queryBy,
-		order: { currentStatusDate: "DESC" },
-		take: take(query),
-		skip: skip(query)
-	});
+	if (query.user_id) {
+		const contexts = await OrderContext.find({ userId: query.user_id });
+		if (contexts.length > 0) {
```
liorama: `if (contexts.length) {`

liorama: but maybe do:
`contexts.length ? q.andWhere("id in (:ids)", { ids: contexts.map(c => c.orderId) }) : q.andWhere("id = 'no-such-id' ");`
Feels nicer when it's all on one line. I don't think it's too much.

liorama: What's `'no-such-id'` btw?

scripts/src/admin/services.ts
```diff
@@ -546,12 +554,12 @@ export async function changeOffer(body: Partial<Offer>, params: { offer_id: stri
...
 	if (body && body.cap) {
-		if (body.cap.total && !isNothing(parseInt(body.cap.total as any, 10)) && parseInt(body.cap.total as any, 10) >= 0) {
-			offer.cap.total = parseInt(body.cap.total as any, 10);
+		if (!isNothing(body.cap.total)  && body.cap.total >= 0) {
+			offer.cap.total = body.cap.total;
 			didChange = true;
 		}
-		if (body.cap.per_user && !isNothing(parseInt(body.cap.per_user as any, 10)) && parseInt(body.cap.per_user as any, 10) >= 0) {
-			offer.cap.per_user = parseInt(body.cap.per_user as any, 10);
+		if (!isNothing(body.cap.per_user) && body.cap.per_user >= 0) {
```
liorama: What's the meaning of `body.cap.per_user = 0` is it no cap or a cap 0 so no offers at all?
If it's no cap then you can do `if (body.cap.per_user) {`

scripts/src/admin/services.ts
```diff
@@ -405,22 +411,23 @@ export async function getApplicationUserData(params: { app_user_id: string, app_
...
 export async function getOrders(params: any, query: Paging & { status?: OpenOrderStatus, user_id?: string, offer_id?: string }): Promise<string> {
-	const queryBy: { offerId?: string, userId?: string, status?: OpenOrderStatus } = {};
+	const q = await Order.queryBuilder("order");
+
 	if (query.offer_id) {
-		queryBy.offerId = query.offer_id;
-	}
-	if (query.user_id) {
-		queryBy.userId = query.user_id;
+		q.andWhere("offer_id = :offer_id", { offer_id: query.offer_id });
```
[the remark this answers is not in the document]
liorama: They are not interchangable https://github.com/typeorm/typeorm/blob/a4dec02cc59d3219a29c7be0322af2253e1452dc/src/query-builder/SelectQueryBuilder.ts#L701

liorama:
Also, make sure to squash all these commits

scripts/src/admin/services.ts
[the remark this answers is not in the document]
doodyparizada: It actually did this on purpose. The problem with `where` is that if you call it twice by mistake, it overrides the previous where. but with `andWhere` this won't happen

scripts/src/admin/services.ts
```diff
@@ -405,22 +411,23 @@ export async function getApplicationUserData(params: { app_user_id: string, app_
 }
 
 export async function getOrders(params: any, query: Paging & { status?: OpenOrderStatus, user_id?: string, offer_id?: string }): Promise<string> {
-	const queryBy: { offerId?: string, userId?: string, status?: OpenOrderStatus } = {};
+	const q = await Order.queryBuilder("order");
```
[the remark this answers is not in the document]
doodyparizada: k

scripts/src/admin/services.ts
doodyparizada: it means no offer will return to any user. 0 is a valid option for cap.

scripts/src/admin/services.ts
doodyparizada: If a user has no orders, an empty list should return. `WHERE id IN ()` is an illegal SQL, so I'm just using an operation that will never return. I could of also just run `WHERE 1 = 0`

PR #168, opened 2018-09-04
ECO-603 Add translation table

liorama:


scripts/src/models/translations.ts
```diff
@@ -0,0 +1,32 @@
+import { BaseEntity, Column, Entity, JoinColumn, ManyToOne, ObjectType, OneToMany, PrimaryColumn } from "typeorm";
+import { register as Register } from "./index";
+import { DeepPartial } from "typeorm/common/DeepPartial";
```
nitzantomer: as we talked about, i prefer that the npm modules be at the top, especially as it is an import from `typeorm` then put it with the other import from there

liorama:
@nitzantomer It was a different task hence different PR, I'll wait with merging it.

nitzantomer:
@liorama 
so that's not good.
either don't split up to such small tasks, or group a few into a PR, because you can't go over something like this without a context.  as an example, again is whether or not you need to have the `OfferTranslations.new` function.

PR #171, opened 2018-09-06
handle user register race condition

doodyparizada:


scripts/src/public/services/users.ts
```diff
@@ -38,15 +39,7 @@ export async function getOrCreateUserCredentials(
...
-	if (!user) {
-		logger.info("creating a new user", { appId, appUserId });
-		// new user
-		user = db.User.new({ appUserId, appId, walletAddress });
-		await user.save();
-		logger.info(`creating stellar wallet for new user ${user.id}: ${user.walletAddress}`);
-		await payment.createWallet(user.walletAddress, user.appId, user.id, logger);
-		metrics.userRegister(true, true);
-	} else {
+	async function handleExistingUser(user: User) {
```
nitzantomer: is there a reason why this function is defined here and not on the module scope?  
it means that this definition is being parsed in every execution of the top function (getOrCreateUserCredentials).

also, it's confusing calling the argument `user` as there's already such a variable two lines above.

either extract this function to the module level or rename this argument.

doodyparizada: I just wanted the scoping, but I'll move it outside

nitzantomer: @doodyparizada 
i understand the motivation behind it as the closure gives you access to other variables.
the issue here is performance, the function is recreated everytime the parent function is executed.

i'm not saying it's wrong, just think of the pros/cons for the reasons and decide, i don't really mind.
what i do mind though is that if you keep it there, change the argument name from `user` as you now have to variables with that name and it is confusing.

doodyparizada: I don't think we should care about the performance of defining the closure on each run of the function - accessing the DB should be much slower (I hope I'm not wrong). I'll change the argument name.

PR #174, opened 2018-09-13
handle both user id and jid

nitzantomer:
# DO NOT MERGE

I'm not sure that we want this in master, maybe just deploy it and let it run until the migration is done?

### DIFF
--- a/scripts/src/public/services/native_offers.ts
+++ b/scripts/src/public/services/native_offers.ts
@@ -61,13 +61,13 @@ export async function validateExternalOrderJWT(jwt: string, user: User, logger: 
 		const decodedUserAppId = decoded.payload.sender.user_jid ? decoded.payload.sender.user_id : null;
 		const decodedUserAppJid = decoded.payload.sender.user_jid ? decoded.payload.sender.user_jid : decoded.payload.sender.user_id;
 
-		logger.info(`[JID_MIGRATION] validateExternalOrderJWT\n\t\tuserAppId:${userAppId}\n\t\tuserAppJid:${userAppJid}\n\t\tdecodedUserAppId:${decodedUserAppId}\n\t\tdecodedUserAppJid:${decodedUserAppJid}`);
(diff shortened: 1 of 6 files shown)


scripts/src/public/services/native_offers.ts
```diff
@@ -40,22 +47,36 @@ export function isPayToUser(jwt: ExternalOrderJWT): jwt is ExternalPayToUserOrde
...
-	if ((decoded.payload.sub === "spend" || decoded.payload.sub === "pay_to_user") &&
-		!!decoded.payload.sender.user_id && decoded.payload.sender.user_id !== appUserId) {
+	// TEMP:JID_MIGRATION
+	if (decoded.payload.iss === "kik") {
+		const userAppId = user.appUserJid ? user.appUserId : null;
+		const userAppJid = user.appUserJid ? user.appUserJid : user.appUserId;
+		const decodedUserAppId = decoded.payload.sender.user_jid ? decoded.payload.sender.user_id : null;
+		const decodedUserAppJid = decoded.payload.sender.user_jid ? decoded.payload.sender.user_jid : decoded.payload.sender.user_id;
+
+		if ((decoded.payload.sub === "spend" || decoded.payload.sub === "pay_to_user")) {
```
doodyparizada: this might fail if kik only changes the register JWT and not the spend. so `userAppId` will be defined but `decodedUserAppId` won't. Should you support this case?

scripts/src/public/services/users.ts
```diff
@@ -30,15 +30,37 @@ function AuthTokenDbToApi(authToken: db.AuthToken, user: db.User, logger: Logger
...
-	appId: string,
+	appUserJid: string | null,
 	walletAddress: string,
 	deviceId: string, logger: LoggerInstance): Promise<AuthToken> {
 
+	const appId = app.id;
+
 	async function handleExistingUser(existingUser: User) {
-		logger.info("found existing user", { appId, appUserId, userId: existingUser.id });
+		logger.info("found existing user", { app, appUserId, userId: existingUser.id });
```
doodyparizada: { appId: app.id, ...

scripts/src/public/services/native_offers.ts
nitzantomer: but if they don't change their spend JWT then they will send the `jid` as the regular id, so the matching will fail in any case...
they have to change all JWTs or none at all

PR #178, opened 2018-09-24
ECO-605 add import treanslations CSV

liorama:


scripts/src/admin/translations.ts
```diff
@@ -150,8 +158,118 @@ export async function getCsvTemplateData() {
...
+	});
+	OfferTranslation.createQueryBuilder().insert().values(dbReadyData).execute();
+}
+
+//  TODO: add validation
+async function processTranslationData(csvData: TranslationData) {
+	const allOfferContents = await OfferContent.find({ select: ["offerId", "content"] } as FindManyOptions<OfferContent>);
+	const allContentTranslations: OffersTranslation = {};
+	csvData.forEach(row => {
+		const [__, csvKey, ___, translation] = row;
```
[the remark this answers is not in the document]
liorama: I think it's clearer this way, since it's more explicit.

scripts/src/admin/translations.ts
```diff
@@ -150,8 +158,136 @@ export async function getCsvTemplateData() {
...
+		const [table, offerId, column, jsonPath] = getCsvKeyElements(csvKey);
+		const offerTranslations = (offerId in allContentTranslations) ?
```
nitzantomer: didn't we say that you're breaking this to a more readable `if () { } else {}` block?

scripts/src/admin/translations.ts
```diff
@@ -150,8 +158,136 @@ export async function getCsvTemplateData() {
...
+	orderTitle: string,
+	content: any /* It's suppose to be offerContent: { content } but that drives TS crazy */
+};
+type Column = "title" | "description" | "orderDescription" | "orderTitle" | "content";
+type Table = "offer" | "offerContent";
+type OffersTranslation = { [index: string]: OfferTranslationData };
+type OffersTranslationRow = {
+	offer: Offer;
+	offerId: string;
+	context: "offer" | "offer_content";
```
nitzantomer: can't you use the `Table` type here? it's pretty much the same

scripts/src/admin/translations.ts
```diff
@@ -150,8 +158,136 @@ export async function getCsvTemplateData() {
...
+			} catch (e) {
+			}
+		}
+		allContentTranslations[offerId] = offerTranslations;
+	});
+	return allContentTranslations;
+}
+
+//  rowOffset is a 0 base index of the row to start with
+export async function processFile(filename: string, languageCode: string, rowOffset = 1) {
```
nitzantomer: doesn't look like this function is used, is it necessary?

scripts/src/admin/translations.ts
```diff
@@ -150,8 +158,136 @@ export async function getCsvTemplateData() {
...
+
+function getOfferContentFromJson(offerContent?: OfferContent) {
+	if (!offerContent) {
+		return {};
+	}
+	return parseContent(offerContent.content);
+}
+
+async function insertIntoDb(data: OffersTranslation, language: string) {
+	const allOffers = await Offer.find({ type: "earn" });
```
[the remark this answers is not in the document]
liorama: Moti says this will be less performant DB wise. So it's a question of micro-optimising DB vs in memory iteration...I'm leaving it as it is....

PR #180, opened 2018-10-07
Lior/eco 609 submit order answer validation supports i18n

liorama:


scripts/src/public/services/offers.ts
```diff
@@ -72,12 +72,10 @@ async function filterOffers(userId: string, app: Application | undefined, logger
...
-	let language: string | null = null;
+	let language: string | false = false;
 	if (acceptsLanguagesFunc) {
-		availableTranslations = await OfferTranslation.createQueryBuilder("translations")
-			.where("translations.language IN (:languages)", { languages: acceptsLanguagesFunc() })
-			.getMany();
-		const availableLanguages = new Set(availableTranslations.map(translation => translation.language));
+		let availableLanguages;
+		[availableLanguages, availableTranslations] = await OfferTranslation.getSupportedLanguages({ languages: acceptsLanguagesFunc() });
 		language = acceptsLanguagesFunc(Array.from(availableLanguages)); // get the most suitable language for the client
```
doodyparizada: availableLanguages is already an Array

scripts/src/public/services/offer_contents.ts
```diff
@@ -1,7 +1,10 @@
 import { LoggerInstance } from "winston";
+import { Request as ExpressRequest } from "express-serve-static-core";
 
 import { isNothing } from "../../utils";
 import * as db from "../../models/offers";
+import { OfferTranslation } from "../../models/translations";
+import { Path } from "../../admin/translations";
```
doodyparizada: I wouldn't put `Path` in admin if you're importing it into the public server. it should probably be the opposite

doodyparizada: just define it here `scripts/src/models/translations.ts`

scripts/src/public/services/offer_contents.ts
```diff
@@ -123,7 +126,7 @@ export function isValid(offerContent: db.OfferContent, form: string | undefined)
 	return typeof answers === "object" && !Array.isArray(answers);
 }
 
-export function sumCorrectQuizAnswers(offerContent: db.OfferContent, form: string | undefined): number {
+export async function sumCorrectQuizAnswers(offerContent: db.OfferContent, form: string | undefined, acceptsLanguagesFunc?: ExpressRequest["acceptsLanguages"]): Promise<number> {
```
doodyparizada: Why doesn't this only pass the actual language that the poll was served in?
after you change the answers to ids, you wouldn't need to know this anyway

doodyparizada: It just feels broken that you might be served the poll in one language and somehow check the results in another one

scripts/src/public/services/offers.ts
```diff
@@ -72,12 +72,10 @@ async function filterOffers(userId: string, app: Application | undefined, logger
 	const offerCounts = await Order.countAllByOffer(userId);
 	const contents = await offerContents.getAllContents();
 	let availableTranslations: OfferTranslation[] = [];
-	let language: string | null = null;
+	let language: string | false = false;
```
doodyparizada: why false?

scripts/src/public/services/offers.ts
```diff
@@ -72,12 +72,10 @@ async function filterOffers(userId: string, app: Application | undefined, logger
...
 	if (acceptsLanguagesFunc) {
-		availableTranslations = await OfferTranslation.createQueryBuilder("translations")
-			.where("translations.language IN (:languages)", { languages: acceptsLanguagesFunc() })
-			.getMany();
-		const availableLanguages = new Set(availableTranslations.map(translation => translation.language));
+		let availableLanguages;
+		[availableLanguages, availableTranslations] = await OfferTranslation.getSupportedLanguages({ languages: acceptsLanguagesFunc() });
```
doodyparizada: The api for `getSupportedLanguages` doesn't feel clean. Why would it return the `availableLanguages` if you can deduce it from the `availableTranslations`

scripts/src/models/translations.ts
```diff
@@ -10,6 +17,30 @@ export class OfferTranslation extends BaseEntity {
...
+		if (languages) {
+			query.where("translations.language IN (:languages)", { languages });
+		}
+		if (offerId) {
+			query.andWhere("translations.offer_id = :offerId", { offerId });
+		}
+		if (paths) {
+			query.andWhere("translations.path IN (:paths)", { paths });
+		}
+		return await query.getMany();
```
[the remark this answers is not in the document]
liorama: No you can't...not in our current version:
https://kinecosystem.slack.com/archives/CAC2GRPU7/p1538304703000100

scripts/src/public/services/offer_contents.ts
liorama: I don't want to add a language column to Order. This will all change, as you said, in a few days so it doesn't matter much.

scripts/src/public/services/offers.ts
liorama: That's what the `acceptsLanguages` function returns when no supported language is found.

scripts/src/public/services/offers.ts
liorama: Helper function. We do this in a few places, DRY.

scripts/src/public/services/offer_contents.ts
doodyparizada: kk, but this makes that task even more important

liorama: Anything that will make us do that task and not postpone it anymore is a good thing.

PR #181, opened 2018-10-09
run create-jwt-keys before building the docker image

doodyparizada:


Makefile
```diff
@@ -48,7 +48,7 @@ push-image:
 pull:
 	docker-compose -f docker-compose.yaml -f docker-compose.deps.yaml pull
 
-up:
+up: db-docker
```
liorama: Why?

doodyparizada: because the internal server fails without having a DB first

PR #186, opened 2018-10-14
Eco 652 cap per DS

doodyparizada:
Currently all digital service user earning is payed from the same Kin account with a shared capping per offer. This prevent us for providing separate funds per each DS.

In this PR, the capping and payment address is configurable per DS via the AppOffer class.

### DIFF
--- a/scripts/src/admin/routes.ts
+++ b/scripts/src/admin/routes.ts
@@ -179,7 +179,7 @@ export function createRoutes(app: Express, pathPrefix?: string) {
 		.get("/orders/:order_id/retry", wrapService(retryOrder))
 		.get("/users/:user_id/retry", wrapService(retryUserWallet))
 		// change data
-		.post("/applications/:app_id//offers/:offer_id", jsonResponse(changeOffer))
+		.post("/applications/:app_id/offers/:offer_id", jsonResponse(changeOffer))
 	;
 
 	app.use("", router);
--- a/scripts/src/admin/routes.ts
+++ b/scripts/src/admin/routes.ts
@@ -179,7 +179,7 @@ export function createRoutes(app: Express, pathPrefix?: string) {
 		.get("/orders/:order_id/retry", wrapService(retryOrder))
 		.get("/users/:user_id/retry", wrapService(retryUserWallet))
 		// change data
-		.post("/offers/:offer_id", jsonResponse(changeOffer))
+		.post("/applications/:app_id//offers/:offer_id", jsonResponse(changeOffer))
 	;
 
 	app.use("", router);
--- a/scripts/src/admin/services.ts
+++ b/scripts/src/admin/services.ts
@@ -1,5 +1,5 @@
-import { Application } from "../models/applications";
-import { Offer, PollAnswer } from "../models/offers";
+import { Application, AppOffer } from "../models/applications";
+import { Cap, Offer, PollAnswer } from "../models/offers";
 import { getManager } from "typeorm";
 import { User } from "../models/users";
 import { OpenOrderStatus, Order, OrderContext } from "../models/orders";
@@ -52,8 +52,6 @@ const OFFER_HEADERS = `<tr>
 <th>total cap</th>
 <th>total per user</th>
 <th>owner</th>
-<th>recipient</th>
-<th>sender</th>
 <th>date</th>
 </tr>`;
 
@@ -402,10 +400,9 @@ export async function getApplicationOffers(params: { app_id: string }, query: Pa
 		throw new Error("no such app: " + params.app_id);
 	}
 
-	const offers = app.offers;
 	let ret = `<table>${OFFER_HEADERS}`;
-	for (const offer of offers) {
-		ret += await offerToHtml(offer);
+	for (const appOffer of app.appOffers) {
+		ret += await offerToHtml(appOffer.offer);
 	}
 	ret += "</table>";
 	return ret;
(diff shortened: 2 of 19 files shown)


scripts/src/create.ts
```diff
@@ -13,7 +13,7 @@ import { init as initModels, close as closeModels } from "./models";
 import { PageType, Poll, Quiz, Tutorial } from "./public/services/offer_contents";
 import { createEarn, createSpend } from "./create_data/offers";
 import { ContentType, Offer } from "./models/offers";
-import { StringMap, Application, ApplicationConfig } from "./models/applications";
+import { StringMap, Application, ApplicationConfig, AppOffer } from "./models/applications";
```
liorama: Where are you using this?

scripts/src/admin/routes.ts
```diff
@@ -179,7 +179,7 @@ export function createRoutes(app: Express, pathPrefix?: string) {
 		.get("/orders/:order_id/retry", wrapService(retryOrder))
 		.get("/users/:user_id/retry", wrapService(retryUserWallet))
 		// change data
-		.post("/offers/:offer_id", jsonResponse(changeOffer))
+		.post("/applications/:app_id//offers/:offer_id", jsonResponse(changeOffer))
```
doodyparizada: two `//`

scripts/src/internal/services.ts
```diff
@@ -194,9 +195,17 @@ export async function paymentFailed(payment: FailedPayment, logger: LoggerInstan
  * register to get callbacks for incoming payments for all the active offers
  */
 export async function initPaymentCallbacks(logger: LoggerInstance): Promise<Watcher> {
-	const offers = await Offer.find<Offer>({ type: "spend" }); // get all active spend offers
+	const appOffers = await AppOffer.find();
+	const apps = await Application.find();
```
liorama: You can use `promise.all()` with these queries as well...

scripts/src/internal/services.ts
```diff
@@ -194,9 +195,17 @@ export async function paymentFailed(payment: FailedPayment, logger: LoggerInstan
...
 	// create a list of unique addresses
-	const addresses = removeDuplicates(offers.map(offer => offer.blockchainData.recipient_address!));
+	const addresses = removeDuplicates(
+		[
+			...appOffers
+			.filter(appOffer => appOffer.offer.type === "spend")
+			.map(appOffer => appOffer.walletAddress),
```
liorama: Add parentheses to make it more readable 
```javascript
	...(appOffers
	.filter(appOffer => appOffer.offer.type === "spend")
	.map(appOffer => appOffer.walletAddress)),
```

scripts/src/public/services/offers.ts
```diff
@@ -90,41 +86,42 @@ async function filterOffers(userId: string, app: Application | undefined, logger
...
-					}
-					const content = contents.get(offer.id);
-					if (!content) {
-						return null;
-					}
-					return offerDbToApi(offer, content, getOfferTranslations(language, offer.id, availableTranslations));
+		appOffers
+			.map(async appOffer => {
+				const offer = appOffer.offer;
+				if ((totalOfferCounts.get(offer.id) || 0) >= appOffer.cap.total) {
```
liorama: We should add a default value to `Map#get`, like in python.

scripts/src/models/applications.ts
```diff
@@ -45,6 +57,59 @@ export class Application extends CreationDateModel {
...
+
+	@ManyToOne(type => Application, app => app.appOffers)
+	@JoinColumn({ name: "applicationsId" })
+	public readonly app!: Application;
+
+	public async didExceedCap(userId: string): Promise<boolean> {
+		const total = await Order.countByOffer(this.offerId);
+
+		if (total >= this.cap.total) {
+			return true;
```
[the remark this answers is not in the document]
doodyparizada: A caching solution is something a bit bigger, I don't want to add it now.
My idea is to manage these totals per offer+app in a cache and never hit the DB for counting.

scripts/src/public/services/offers.ts
doodyparizada: ok. How do you suggest?

scripts/src/models/orders.ts
```diff
@@ -22,6 +22,7 @@ import { generateId, IdPrefix } from "../utils";
 import { User } from "./users";
 import { BlockchainData, OfferType, OrderValue } from "./offers";
 import { CreationDateModel, initializers as Initializers, register as Register } from "./index";
+import { Exception } from "winston";
```
doodyparizada: why this import?

doodyparizada:
## migration
```sql
ALTER TABLE applications_offers_offers ADD column cap text;
ALTER TABLE applications_offers_offers ADD column wallet_address varchar;

UPDATE applications_offers_offers as ao SET cap = o.cap, wallet_address = CASE WHEN o.type='spend' THEN o.blockchain_data::json->'recipient_address' ELSE o.blockchain_data::json->'sender_address' END  FROM offers o WHERE "offersId" = o.id;
```
Then remove old columns and make these columns not null;

scripts/src/admin/services.ts
```diff
@@ -262,7 +261,28 @@ async function appToHtml(app: Application): Promise<string> {
 </tr>`;
 }
 
-async function offerToHtml(offer: Offer): Promise<string> {
+async function offerToHtml(offer: Offer, appOffer?: AppOffer): Promise<string> {
```
YohayBar: @doodyparizada did you add/adjust unit tests?

doodyparizada: This is the admin page. Admin isn't currently part of the unit tests. It could be when we migrate it to an API service instead of server-side rendering.

doodyparizada:
added description

scripts/src/public/services/offers.ts
liorama: Subclass or Proxy

PR #187, opened 2018-10-17
ECO-561 remove need of activate

doodyparizada:
Remove the need to activate the account before using the SDK.

### DIFF
--- a/scripts/src/public/routes/index.ts
+++ b/scripts/src/public/routes/index.ts
@@ -50,7 +50,7 @@ function proxyOverRouter(router: express.Router, proxy: ExtendedRouter, obj: any
 
 const AUTHENTICATED_METHODS = ["get", "delete", "post", "put", "patch"];
 
-enum AuthScopes { TOS }
+enum AuthScopes { }
 
 function Router(): ExtendedRouter {
 	const router = express.Router() as ExtendedRouter;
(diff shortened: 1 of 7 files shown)


scripts/src/public/routes/index.ts
```diff
@@ -63,8 +63,10 @@ function Router(): ExtendedRouter {
 						return (target as any)[name](path, async (req: express.Request, res: express.Response, next: express.NextFunction) => {
 							const token = await authenticate(req);
 							const user = await db.User.findOneById(token.userId);
-							// XXX scopes should be per token and should not consider user data
-							if (scopes.includes(AuthScopes.TOS) && (!user || !user.activated || token.createdDate < user.activatedDate!)) {
+
+							if (!user) {
+								// This error now defines an inconsistent state in the DB where a token exists but not user is found
+								// This should never happen as the token.user_id is a foreign key to the users table
 								throw TOSMissingOrOldToken();
```
YohayBar: maybe change the exception to UserWasNotFound or something like that

scripts/src/public/routes/index.ts
```diff
@@ -63,8 +63,10 @@ function Router(): ExtendedRouter {
 						return (target as any)[name](path, async (req: express.Request, res: express.Response, next: express.NextFunction) => {
 							const token = await authenticate(req);
 							const user = await db.User.findOneById(token.userId);
-							// XXX scopes should be per token and should not consider user data
-							if (scopes.includes(AuthScopes.TOS) && (!user || !user.activated || token.createdDate < user.activatedDate!)) {
```
doodyparizada: remove the extra token from the activate method

YohayBar:
@doodyparizada Do you also change the auth token and remove activated or you set activated true when returning the the first token?

scripts/src/public/routes/index.ts
liorama: 👍

liorama:
You need to remove the `didNotApproveTOS` system test

PR #192, opened 2018-10-21
add app_id to client error metrics

doodyparizada:
remove client headers and added app_id to metrics, so we can separate errors per app

### DIFF
--- a/scripts/src/metrics.ts
+++ b/scripts/src/metrics.ts
@@ -12,50 +12,50 @@ export function destruct() {
 }
 
 export function userRegister(newUser: boolean, newWallet: boolean) {
-	statsd.increment("user_register", 1, undefined, { new_user: newUser.toString(), new_wallet: newWallet.toString() });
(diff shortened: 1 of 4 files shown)


scripts/src/metrics.ts
```diff
@@ -45,9 +45,9 @@ export function offersReturned(numOffers: number, appId: string) {
 	statsd.histogram("offers_returned", numOffers, undefined, { app_id: appId });
 }
 
-export function reportClientError(error: MarketplaceError, headers: { [name: string]: string }) {
-	const data = Object.assign({ status: error.status.toString(), title: error.title, code: error.code }, headers);
-	statsd.increment("client_error", 1, undefined, data);
+export function reportClientError(error: MarketplaceError, appId: string) {
+	statsd.increment("client_error", 1, undefined,
```
liorama: Why `undefined` and not `null`? null signifies no value, undefined means this argument wasn't processed/touched by the code, which is indicative of a bug.

PR #199, opened 2018-11-01
Nitzan/eco 678/order nonce

nitzantomer:


scripts/src/mock_client.ts
```diff
@@ -27,18 +27,32 @@ const API_KEY = process.env.API_KEY || Application.SAMPLE_API_KEY;  // get this
...
 class SampleAppClient {
+	private static createJwtRequestUrl(base: string, params: { [key: string]: string | undefined | null }) {
+		const url = (JWT_SERVICE_BASE + "/" + base);
+		return url + "?" + Object.entries(params).reduce((previous: string, [key, value]) => {
+			if (value === null || value === undefined) {
+				return previous;
+			}
+
+			return previous + `&${ key }=${ value }`;
+		}, "").substring(1);
```
doodyparizada: I think axios does this for you

doodyparizada:
```javascript
axios.get('/user', {
    params: {
      ID: 12345
    }
  })
```

scripts/src/mock_client.ts
```diff
@@ -571,6 +617,7 @@ async function main() {
 	await didNotApproveTOS();
 	await testRegisterNewUser();
 	await tryToNativeSpendTwice();
+	await tryToNativeSpendTwiceWithNonce();
```
doodyparizada: nice!

scripts/src/models/orders.ts
```diff
@@ -310,12 +321,15 @@ export type ExternalOrderFactory = OrderFactory & {
 function extendedOrder(origin: OrderOrigin): (typeof Order) & OrderFactory {
 	return Object.assign({}, Order, {
 		"new"(data?: DeepPartial<Order>, ...context: Array<DeepPartial<OrderContext>>): Order {
-			data = Object.assign({}, data, { origin });
+			data = Object.assign(
+				{ nonce: Order.DEFAULT_NONCE },
+				data,
+				{ origin });
```
doodyparizada: OMG don't know how to read this

scripts/src/public/services/orders.ts
```diff
@@ -223,8 +223,9 @@ async function createNormalSpendExternalOrder(sender: User, jwt: ExternalSpendOr
 
 export async function createExternalOrder(jwt: string, user: User, logger: LoggerInstance): Promise<OpenOrder> {
 	const payload = await validateExternalOrderJWT(jwt, user.appUserId, logger);
+	const nonce = payload.nonce || db.Order.DEFAULT_NONCE;
 
-	let order = await db.Order.findBy(payload.offer.id, user.id);
+	let order = await db.Order.findBy({ offerId: payload.offer.id, userId: user.id, nonce });
```
doodyparizada: There is a bug here that you don't order the Orders by date, so if you failed once the findBy might always return the failed order instead of the latest one.
I think you should add `order by date desc` to `db.Order.findBy` function

scripts/src/models/orders.ts
nitzantomer: what's the problem?
start with the default object with nonce, then update and override with `data` and then add the origin.

scripts/src/public/services/orders.ts
nitzantomer: nice catch.  added

scripts/src/mock_client.ts
```diff
@@ -586,6 +629,7 @@ async function main() {
 	await didNotApproveTOS();
 	await testRegisterNewUser();
 	await tryToNativeSpendTwice();
+	await tryToNativeSpendTwiceWithNonce();*/
```
doodyparizada: return this

PR #201, opened 2018-11-13
fixed the nonce bug

nitzantomer:
there wasn't an option to not use the nonce, the previous code assumed that if no nonce was sent then the 'nonce' property wouldn't exist, but in reality it was there but with the value as undefined.
the tests did not catch it because the JWT service we used would add it's own nonce default value, that was fixed in the JWT service repo and the updated image is now used here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "sqlite3": "^4.0.2",
     "tslint": "^5.9.1",
     "tslint-eslint-rules": "^4.1.1",
-    "typescript": "^2.7.2"
+    "typescript": "^2.8"
   },
   "scripts": {
     "clean": "rimraf scripts/bin",
--- a/scripts/src/models/orders.ts
+++ b/scripts/src/models/orders.ts
@@ -17,7 +17,7 @@ import {
 } from "typeorm";
 
 import { ApiError } from "../errors";
-import { generateId, IdPrefix, Mutable } from "../utils";
+import { generateId, IdPrefix, Mutable, isNothing } from "../utils";
 
 import { User } from "./users";
 import { BlockchainData, OfferType, OrderValue } from "./offers";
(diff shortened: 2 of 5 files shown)


scripts/src/models/orders.ts
```diff
@@ -323,11 +323,15 @@ export type ExternalOrderFactory = OrderFactory & {
...
 function extendedOrder(origin: OrderOrigin): (typeof Order) & OrderFactory {
 	return Object.assign({}, Order, {
-		"new"(data?: DeepPartial<Order>, ...context: Array<DeepPartial<OrderContext>>): Order {
+		"new"(data?: Mutable<DeepPartial<Order>>, ...context: Array<DeepPartial<OrderContext>>): Order {
 			data = Object.assign(
-				{ nonce: Order.DEFAULT_NONCE },
 				data,
 				{ origin });
+
+			if (typeof data.nonce !== "string") {
```
doodyparizada: why wouldn't it be string?

doodyparizada: isNothing

package.json
```diff
@@ -59,7 +59,7 @@
     "sqlite3": "^4.0.2",
     "tslint": "^5.9.1",
     "tslint-eslint-rules": "^4.1.1",
-    "typescript": "^2.7.2"
+    "typescript": "^2.8"
```
liorama: Please put this into its own commit or at least add a comment about this in the commit summary.

scripts/src/models/orders.ts
liorama: I would check if the property exists vs checking it's value, eg `data.hasOwnProperty('nounce')` or `'nounce' in data` if you want to live on the edge (I think it'll be ok).

scripts/src/models/orders.ts
```diff
@@ -325,9 +325,13 @@ function extendedOrder(origin: OrderOrigin): (typeof Order) & OrderFactory {
 	return Object.assign({}, Order, {
 		"new"(data?: DeepPartial<Order>, ...context: Array<DeepPartial<OrderContext>>): Order {
 			data = Object.assign(
-				{ nonce: Order.DEFAULT_NONCE },
 				data,
 				{ origin });
+
+			if (isNothing(data.nonce)) {
```
liorama: I would check if the property exists vs checking it's value, eg `data.hasOwnProperty('nounce')` or `'nounce' in data` if you want to live on the edge (I think it'll be ok). (cc @doodyparizada)

scripts/src/models/orders.ts
```diff
@@ -325,9 +325,13 @@ function extendedOrder(origin: OrderOrigin): (typeof Order) & OrderFactory {
...
+				(data as Mutable<Order>).nonce = Order.DEFAULT_NONCE;
```
YohayBar: you need to make sure no one will alter Order.DEFAULT_NONCE in the future - might be an overkill but simple comment can help

PR #208, opened 2018-11-19
Caching AppOffers, OfferContent and translations

liorama:


scripts/src/models/translations.ts
```diff
@@ -18,9 +20,15 @@ export class OfferTranslation extends BaseEntity {
 
 	public static async getTranslations(criteria: GetTranslationsCriteria = {}): Promise<OfferTranslation[]> {
 		//  todo add cache
+		const cacheKey = JSON.stringify(criteria);
```
doodyparizada: stringify will give you different dumps depending on the structure in memory, no?

liorama: mmm there is a chance it will change the order, ok I'll change it just in case.

scripts/src/models/translations.ts
```diff
@@ -18,9 +20,15 @@ export class OfferTranslation extends BaseEntity {
 
 	public static async getTranslations(criteria: GetTranslationsCriteria = {}): Promise<OfferTranslation[]> {
 		//  todo add cache
+		const cacheKey = `${criteria.languages}:${criteria.offerId}:${criteria.paths}`;
```
doodyparizada: should be sorted(criteria.languages)

liorama: The language order has meaning, language priority.

liorama: paths can be sorted, but maybe let's deploy it this way, so it'll hit the DB once or twice more instead of fetching from cache

doodyparizada:
looks good if passes tests I'll deploy

PR #209, opened 2018-11-19
cache tokens

doodyparizada:


scripts/src/public/auth.ts
```diff
@@ -1,17 +1,40 @@
...
+	const authToken = await db.AuthToken.findOneById(token);
+	if (!authToken) {
+		throw InvalidToken(token);
+	}
+
+	return authToken;
+}
+
+export async function authenticateAndGetUser(token: string): Promise<[db.AuthToken, db.User]> {
+	const redis = getRedisClient();
```
liorama: Can this ever fail? should it be in a try-catch?

PR #210, opened 2018-11-20
metrics - added app_id for all metrics.ts methods

Doc999tor:


.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
```
nitzantomer: please add this file to the .gitignore

docker-compose.dev.yaml
```diff
@@ -4,7 +4,7 @@ services:
 
   marketplace-server:
     image: kinecosystem/marketplace-server:latest
-    command: ./tests/wait-for marketplace-internal:3000 -- npm run start
+    #command: ./tests/wait-for marketplace-internal:3000 -- npm run start
```
nitzantomer: why are these commented out?

marketplace.code-workspace
```diff
@@ -0,0 +1,8 @@
+{
```
nitzantomer: also, add to .gitignore

docker-compose.dev.yaml
doodyparizada: @nitzantomer `wait-for` didn't work on @Doc999tor 's machine (Windows). This isn't a solution though. It can be a temp fix for windows devs on their local machines until we find why wait-for failed / replace it with a different impl.

@Doc999tor remove this for the PR and leave it locally only on your machine

scripts/src/internal/services.ts
```diff
@@ -170,7 +170,7 @@ export async function paymentComplete(payment: CompletedPayment, logger: LoggerI
 	await order.save();
 
 	order.forEachContext(context => {
-		metrics.completeOrder(context.type, order.offerId, prevStatus, (order.currentStatusDate.getTime() - prevStatusDate.getTime()) / 1000);
+		metrics.completeOrder(context.type, order.offerId, prevStatus, (order.currentStatusDate.getTime() - prevStatusDate.getTime()) / 1000, payment.app_id);
```
doodyparizada: (reminder to self: I think there should be a single metric here with type=p2p instead of a separate earn/spend)

scripts/src/metrics.ts
```diff
@@ -54,12 +55,12 @@ export function reportClientError(error: MarketplaceError, appId: string) {
...
 
-export function reportServerError(method: string, path: string) {
-	statsd.increment("server_error", 1, { method, path });
+export function reportServerError(method: string, path: string, appId: string) {
+	statsd.increment("server_error", 1, { method, path, app_id: appId });
 }
 
-export function reportProcessAbort(reason: string = "") {
-	statsd.increment("process_abort", 1, { system: "exit", reason });
+export function reportProcessAbort(reason: string = "", appId: string = "") {
```
doodyparizada: not sure app_id is needed here

scripts/src/middleware.ts
```diff
@@ -83,7 +83,8 @@ export const reportMetrics = function(req: express.Request, res: express.Respons
 
 	res.on("finish", () => {
 		const path = req.route ? req.route.path : "unknown";
-		metrics.timeRequest(performance.now() - t, req.method, path);
+		const appId = req.context!.user!.appId;
```
doodyparizada: you won't always have the app_id here. Any request that isn't authorized doesn't have the context. req.context will be undefined and undefined.user will raise an exception

scripts/src/middleware.ts
```diff
@@ -116,7 +117,7 @@ function clientErrorHandler(error: MarketplaceError, req: express.Request, res:
 
 function serverErrorHandler(error: any, req: express.Request, res: express.Response) {
 	const log = req.logger || logger;
-	metrics.reportServerError(req.method, req.url);
+	metrics.reportServerError(req.method, req.url, req.context && req.context.user ? req.context.user.appId : "");
```
doodyparizada: put this same logic above---^ to make sure appId exists or return ""

PR #211, opened 2018-11-22
increase pending time to avoid double spend

doodyparizada:


scripts/src/models/orders.ts
```diff
@@ -441,10 +443,10 @@ class OrderImpl extends CreationDateModel implements Order {
 		this.currentStatusDate = new Date();
 		switch (this.status) {
 			case "pending":
-				this.expirationDate = moment(this.currentStatusDate).add(45, "seconds").toDate();
+				this.expirationDate = moment(this.currentStatusDate).add(PENDING_SECS, "seconds").toDate();
```
YohayBar: need to make sure doesn't affect client behavior

PR #214, opened 2018-11-26
sumCorrectQuizAnswers logic adapted to work with answers indices, moc…

Doc999tor:
…k_client is adapted as well

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "restart": "npm-run-all build start",
     "transpile-tests": "tsc -p tests",
     "test-system": "node scripts/bin/mock_client",
-    "test": "jest --forceExit --runInBand"
+    "test": "jest --forceExit --runInBand --detectOpenHandles"
   },
   "author": "",
   "license": "ISC",
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "@types/winston": "^2.3.7",
     "npm-run-all": "^4.0.1",
     "rimraf": "2.6.1",
-    "sqlite3": "^4.0.2",
+    "sqlite3": "^4.0.4",
     "tslint": "^5.9.1",
     "tslint-eslint-rules": "^4.1.1",
     "typescript": "^2.8"
--- a/package.json
+++ b/package.json
@@ -47,8 +47,8 @@
     "winston": "^2.4.0"
   },
   "devDependencies": {
-    "@types/expect": "^1.20.3",
     "@types/body-parser": "^1.16.8",
+    "@types/expect": "^1.20.3",
     "@types/express": "^4.11.0",
     "@types/jsonwebtoken": "^7.2.5",
     "@types/redlock": "^3.0.1",
@@ -56,7 +56,7 @@
     "@types/winston": "^2.3.7",
     "npm-run-all": "^4.0.1",
     "rimraf": "2.6.1",
-    "sqlite3": "^4.0.2",
+    "sqlite3": "^4.0.4",
     "tslint": "^5.9.1",
     "tslint-eslint-rules": "^4.1.1",
     "typescript": "^2.8"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -47,16 +47,16 @@
...
+    "@types/expect": "^1.20.3",
     "@types/express": "^4.11.0",
     "@types/jsonwebtoken": "^7.2.5",
     "@types/redlock": "^3.0.1",
     "@types/sequelize": "^4.27.4",
     "@types/winston": "^2.3.7",
     "npm-run-all": "^4.0.1",
     "rimraf": "2.6.1",
-    "sqlite3": "^4.0.2",
+    "sqlite3": "^4.0.4",
```
liorama: All dependency changes need to be in their own commit (With a clear commit summary), reason being that if we have a production crisis and need to rollback we now what commit does what.

package.json
```diff
@@ -73,7 +73,7 @@
     "restart": "npm-run-all build start",
     "transpile-tests": "tsc -p tests",
     "test-system": "node scripts/bin/mock_client",
-    "test": "jest --forceExit --runInBand --detectOpenHandles"
+    "test": "jest --forceExit --runInBand"
```
liorama: why?

scripts/src/public/services/offer_contents.ts
```diff
@@ -153,17 +153,16 @@ export async function sumCorrectQuizAnswers(offerContent: db.OfferContent, form:
...
-			if (answerIndex === p.rightAnswer) {
-				amountSum += p.amount;
-			}
-		}
-	}
+			if (answers[page.question.id] === page.rightAnswer) {
+				return sum + page.amount;
+			} else { return sum; }
+		} else { return sum; }
+	}, 0);
```
liorama: Our style guide is to not have one line blocks, but anyway instead of the two `else` clauses just add a ```return sum;``` on this line, just before the end the function so:
```
			if (answers[page.question.id] === page.rightAnswer) {
				return sum + page.amount;
			}
		} 
		return sum; 
	}, 0);
```

scripts/src/public/services/offer_contents.ts
```diff
@@ -153,17 +153,16 @@ export async function sumCorrectQuizAnswers(offerContent: db.OfferContent, form:
 	}
 
 	const quiz: Quiz = JSON.parse(translatedContent || offerContent.content);  // this might fail if not valid json without replaceTemplateVars
-	let amountSum = 0;
+	// let amountSum = 0;
 
-	for (const page of quiz.pages) {
+	const amountSum = quiz.pages.reduce((sum: number, page: QuizPage | SuccessBasedThankYouPage): number => {
```
liorama: I don't mind this change but I'm pretty sure `for-of` is more performant than `Array.prototype.reduce` and I'm really not sure this is more readable. What was your rational for making the change?

package.json
nitzantomer: @liorama 
it sounds good, if i was aware of this flag i would have added it to begin with, but i would also like to know the reason it was added

scripts/src/public/services/offer_contents.ts
nitzantomer: maybe we can add it to our tslint configuration?

package.json
liorama: He removed it. This flag is good for when jest has problems closing. If it doesn't than it should be benign to have it there (worst case, we get a warning sometimes but that's it) on the other hand if do run into a build stuck on jest not closing I prefer having this there so we don't have to push a commit=wait for the build just to see the log, but maybe I'm missing something, hence the comment :)

scripts/src/public/services/offer_contents.ts
liorama: Sure

PR #215, opened 2018-11-27
implement clustering in nodejs

doodyparizada:
I did some tests locally and it works, although the server takes a while to warm up

### DIFF
--- a/scripts/src/public/index.ts
+++ b/scripts/src/public/index.ts
@@ -1,20 +1,41 @@
-import * as http from "http";
+// for cluster master
+import * as cluster from "cluster";
+import * as os from "os";
 
+// for cluster workers
+import * as http from "http";
 import { getConfig } from "./config";
-import { initLogger } from "../logging";
 
 const config = getConfig();
 import { app, init } from "./app";
 
 import { onError, onListening } from "../server";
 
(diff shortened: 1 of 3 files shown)


scripts/src/public/index.ts
```diff
@@ -1,20 +1,41 @@
...
-const server = http.createServer(app);
+if (cluster.isMaster) {
+	// Count the machine's CPUs
+	const cpuCount = os.cpus().length;
+
+	// Create a worker for each CPU
+	os.cpus().forEach(() => cluster.fork());
+
+	// Listen for dying workers
+	cluster.on("exit", function(worker) {
```
nitzantomer: change to arrow function, and remove the empty lines

scripts/src/public/index.ts
```diff
@@ -1,20 +1,41 @@
...
+		cluster.fork();
+
+	});
+} else {
+	init().then(() => {
+		const server = http.createServer(app);
+		server.listen(config.port);
+		server.on("error", onError);
+		server.on("listening", onListening(server));
+		console.log(`Worker ${cluster.worker.id} running`);
```
nitzantomer: remove empty line

PR #216, opened 2018-11-28
prepared statement changed from hardcoded name to dynamic

Doc999tor:


scripts/src/models/orders.ts
```diff
@@ -59,9 +59,9 @@ function updateQueryWithFilter(query: SelectQueryBuilder<any>, name: string, val
 	const fieldName = alias ? `${ alias }.${ name }` : name;
 
 	if (value.startsWith("!")) {
-		query.andWhere(`${ fieldName } != :value`, { value: value.substring(1) });
+		query.andWhere(`${ fieldName } != :${ fieldName }`, { [fieldName]: value.substring(1) });
```
nitzantomer: i find this to be pretty confusing, maybe:
```
const fieldValue = `${ fieldName }_value`;
query.andWhere(`${ fieldName } != :${ fieldValue }`, { [fieldValue]: value.substring(1) });
```

doodyparizada: I don't necessarily agree.

nitzantomer: @doodyparizada 
on what part?

PR #220, opened 2018-12-09
Victor/eco 837/rate limit register

Doc999tor:


doodyparizada:
Try to run rebases from the master instead of merge. It creates a better history.
Also seems like there is still some issue with the config

scripts/src/errors.ts
```diff
@@ -239,3 +243,13 @@ export function BlockchainError(message?: string) {
 export function TransactionTimeout() {
 	return TransactionFailed(CODES.TransactionFailed.TransactionTimeout, "Transaction Timeout");
 }
+
+function TooManyRequests(index: number, message: string): MarketplaceError {
+	return new MarketplaceError(429, index, "Too Many Requests", message);
+}
```
nitzantomer: add an empty line between functions

scripts/src/models/applications.ts
```diff
@@ -21,6 +21,13 @@ export type SignInType = "jwt" | "whitelist";
 export type ApplicationConfig = {
 	max_user_wallets: number | null;
 	sign_in_types: SignInType[];
+	limits: {
+		hourly_registration: number,
```
nitzantomer: types should end with a semicolon instead of a comma

scripts/src/public/config.ts
```diff
@@ -8,6 +8,13 @@ export interface Config extends BaseConfig {
 	max_daily_earn_offers: null | number; // null marks no limit
 	internal_service: string;
 	sign_in_types: string[];
+	limits: {
+		hourly_registration: number,
```
nitzantomer: same: `;` instead of `,`

scripts/src/utils/RateLimit.ts
```diff
@@ -0,0 +1,53 @@
+import * as moment from "moment";
+import { getRedisClient, RedisAsyncClient } from "../redis";
+
+export class RateLimit {
+	private bucketPrefix: string;
```
nitzantomer: seems like the first 3 can be `readonly`

scripts/src/utils/RateLimit.ts
```diff
@@ -0,0 +1,53 @@
...
+
+	public async checkRate(): Promise<boolean> {
+		return await this.check(1);
+	}
+
+	public async checkAmount(amount: number): Promise<boolean> {
+		return await this.check(amount);
+	}
+
+	private async check(step: number) {
```
nitzantomer: Why not:
```
public async check(step?: number) {
    if (step === undefined) {
        step = 1;
    }
}
```
?

scripts/src/utils/utils.ts
```diff
@@ -163,3 +159,28 @@ export function readUTCDate(date: string | Date): Date {
 export function getAppIdFromRequest(req: RequestWithContext): string {
 	return req.context && req.context.user ? req.context.user.appId : "";
 }
+
+export async function throwOnRateLimit(appId: string, type: string, limit: number, duration: Duration) {
```
nitzantomer: why is this one named differently?
that is, why does it start with `throwOn` instead of `check` like the other two?
(i don't care that much, just wondering)

tests/src/services/users.spec.ts
```diff
@@ -24,6 +24,17 @@ describe("api tests for /users", async () => {
 		await metrics.destruct();
 	});
 
+	// test('register user', async () => {
```
nitzantomer: did you mean to push this?

scripts/src/utils/RateLimit.ts
Doc999tor: For semantic reasons only: separate `checkRate` and `checkAmount` names are more clear, inside it encapsulates the generic `check` method call

scripts/src/models/applications.ts
Doc999tor: TS supports commas as types separator in object literals
[https://github.com/facebookarchive/esprima/pull/107](https://github.com/facebookarchive/esprima/pull/107)

nitzantomer: it supports them, but that's not our convention.
type definitions should end with semi-colons

.gitignore
```diff
@@ -1,9 +1,11 @@
 .idea
-scripts/bin
 node_modules
+scripts/bin
+scripts/src/**/*.d.ts
```
nitzantomer: these shouldn't be here.
your ide/env should not produce these files to begin with.
this is true to `.d.ts`, `.js` and `.js.map` files but only in the `scripts/src` directory, the config for the tests is different

scripts/src/utils/RateLimit.ts
```diff
@@ -0,0 +1,74 @@
...
+		this.bucketPrefix = bucketPrefix;
+		this.rateLimitValue = rateLimitValue;
+		this.windowSize = windowSizeMomentObject.asSeconds();
+		this.ttl = this.ttl * this.ttlWindowRatio;
+		this.redis = getRedisClient();
+	}
+
+	public async checkRate(): Promise<boolean> {
+		return await this.check(1);
+	}
```
nitzantomer: add a line break between the methods

scripts/src/utils/RateLimit.ts
nitzantomer: i'm not a fan of this approach, if they are doing the same thing then they should be the same.
but i don't care that much

scripts/src/utils/RateLimit.ts
```diff
@@ -0,0 +1,74 @@
...
+
+		const rateSum: number = (await this.redis.async.mget(...windowKeys))
+			.filter((val: string) => val)
+			.reduce((sum: number, val: string) => sum + Number(val), 0);
+
+		return rateSum >= this.rateLimitValue;
+	}
+}
+
+export async function throwOnRateLimit(appId: string, type: string, limit: number, duration: moment.Duration) {
```
nitzantomer: maybe use `assert` for the name?

scripts/src/utils/RateLimit.ts
Doc999tor: That's the point: they don't do same things, in the code it looks similar but the purpose is different

scripts/src/utils/RateLimit.ts
```diff
@@ -0,0 +1,75 @@
...
+
+	private async check(step: number) {
+		const bucketSize = Math.max(this.windowSize / 60, 1); // resolution
+		const currentTimestampSeconds = Math.floor((Date.now() / 1000) / bucketSize) * bucketSize; // int(currentTimestampSeconds / 60) * 60
+		const currentBucketName = this.bucketPrefix + currentTimestampSeconds;
+
+		await this.redis.async.incrby(currentBucketName, step);
+		this.redis.expire(currentBucketName, this.ttl);
+
+		const windowKeys: string[] = [currentTimestampSeconds.toString()];
```
doodyparizada: `const windowKeys: string[] = []`
I don't understand why you need the first line. Plus it's wrong because it's missing the prefix

scripts/src/utils/RateLimit.ts
```diff
@@ -0,0 +1,75 @@
...
+		for (let i = 0; i < this.windowSize; i += bucketSize) {
+			windowKeys.push(this.bucketPrefix + (currentTimestampSeconds - i));
+		}
+
+		const rateSum: number = (await this.redis.async.mget(...windowKeys))
+			.filter((val: string) => val)
```
doodyparizada: add comment that this removes nulls for non existent buckets

scripts/src/utils/RateLimit.ts
```diff
@@ -0,0 +1,75 @@
...
+			.reduce((sum: number, val: string) => sum + Number(val), 0);
+
+		return rateSum >= this.rateLimitValue;
```
doodyparizada: I'd just check `>`. For example if you have something that happens only once an hour (a limit of 1 an hour). Using this code it would return true on the first go as rateSum includes the current bucket too

scripts/src/models/applications.ts
```diff
@@ -16,11 +16,14 @@ import { Cap, Offer, OfferType } from "./offers";
...
 import { Order, OrderContext } from "./orders";
 
+import { LimitConfig } from "../config";
+
 export type StringMap = { [key: string]: string; };  // key => value pairs
 export type SignInType = "jwt" | "whitelist";
 export type ApplicationConfig = {
 	max_user_wallets: number | null;
 	sign_in_types: SignInType[];
+	limits: LimitConfig
```
nitzantomer: semicolon

scripts/src/public/services/orders.ts
```diff
@@ -103,6 +106,13 @@ export async function changeOrder(orderId: string, userId: string, change: Parti
 }
 
 async function createOrder(appOffer: AppOffer, user: User, orderTranslations = {} as OrderTranslations) {
+	const app = (await Application.findOneById(user.appId))!;
+	if (appOffer.offer.type === "earn") {
+		throwAppEarnLimit(app.id, "total_earn", app.config.limits.minute_total_earn, moment.duration({ minutes: 1 }), appOffer.offer.amount);
```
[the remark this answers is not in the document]
Doc999tor: Renamed

scripts/src/utils/RateLimit.ts
```diff
@@ -0,0 +1,75 @@
...
+	private readonly redis: RedisAsyncClient;
+	private windowSize: number = 0;
+	private ttl: number = 0;
+	private readonly ttlWindowRatio: number = 10;
+
+	constructor(bucketPrefix: string, rateLimitValue: number, windowSizeMomentObject: moment.Duration) {
+		this.bucketPrefix = bucketPrefix;
+		this.rateLimitValue = rateLimitValue;
+		this.windowSize = windowSizeMomentObject.asSeconds();
+		this.ttl = this.ttl * this.ttlWindowRatio;
```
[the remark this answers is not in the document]
Doc999tor: `this.ttl` gets initial value, with initial value TS doesn't cry on possible undefined type for this.ttl

scripts/src/utils/RateLimit.ts
Doc999tor: Prefix added, missed it :(
I need windowKeys for mget asking: it stores all possible keys

scripts/src/utils/RateLimit.ts
Doc999tor: add

scripts/src/utils/RateLimit.ts
doodyparizada: You don't need the first element prepopulated

scripts/src/utils/RateLimit.ts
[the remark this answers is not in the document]
doodyparizada: Where? it's not in the ctor.

scripts/src/utils/utils.ts
```diff
@@ -3,7 +3,10 @@ import { join } from "path";
 import * as fs from "fs";
 
 import { Express } from "express";
-import { Context } from "./public/routes";
+import { Context } from "../public/routes";
+
+import { path } from "./path";
```
doodyparizada: Do you need it here?

scripts/src/utils/RateLimit.ts
Doc999tor: Fixed

scripts/src/utils/utils.ts
Doc999tor: It seems like fix of previous mistake. Context type is used in row 11

scripts/src/adaptTranslationCsv.ts
```diff
@@ -3,7 +3,7 @@ import { ExportToCsv, Options as ExportCsvOptions } from "export-to-csv";
```
doodyparizada: just noticed the filename is camel case and not snake case @liorama

scripts/src/manage_db_data.ts~HEAD
```diff
@@ -0,0 +1,338 @@
+/**
```
doodyparizada: I think this isn't supposed to be here @Doc999tor

scripts/src/public/services/orders.ts
```diff
@@ -103,6 +106,13 @@ export async function changeOrder(orderId: string, userId: string, change: Parti
 }
 
 async function createOrder(appOffer: AppOffer, user: User, orderTranslations = {} as OrderTranslations) {
+	const app = (await Application.findOneById(user.appId))!;
+	if (appOffer.offer.type === "earn") {
+		throwOnAppEarnLimit(app.id, "total_earn", app.config.limits.minute_total_earn, moment.duration({ minutes: 1 }), appOffer.offer.amount);
+		throwOnAppEarnLimit(app.id, "total_earn", app.config.limits.hourly_total_earn, moment.duration({ hours: 1 }), appOffer.offer.amount);
+		throwOnUserEarnLimit(user.id, "user_earn", app.config.limits.hourly_user_earn, moment.duration({ hours: 1 }), appOffer.offer.amount);
```
doodyparizada: didn't we want this on a daily limit of 500?

scripts/src/public/services/orders.ts
```diff
@@ -188,6 +198,8 @@ async function createP2PExternalOrder(sender: User, jwt: ExternalPayToUserOrderJ
 async function createNormalEarnExternalOrder(recipient: User, jwt: ExternalEarnOrderJWT) {
 	const app = await Application.findOneById(recipient.appId);
 
+	throwOnUserEarnLimit(recipient.id, "total_earn", app!.config.limits.hourly_user_earn, moment.duration({ hours: 1 }), jwt.offer.amount);
```
doodyparizada: 1. title should be "user_earn".
2. I think that as a util function, `throwOnUserEarnLimit`, should put this string by itself instead of receiving it as an input. - but lets schedule this to a different PR

scripts/src/utils/RateLimit.ts
```diff
@@ -0,0 +1,101 @@
...
+import { TooManyRegistrations, TooMuchEarnOrdered } from "../errors";
+
+class RateLimit {
+	private readonly bucketPrefix: string;
+	private readonly rateLimitValue: number;
+	private readonly redis: RedisAsyncClient;
+	private windowSize: number = 0;
+	private bucketSize: number = 0;
+	private currentTimestampSeconds: number = 0;
+	private ttl: number = 172800; // two days in seconds
```
doodyparizada: maybe use `moment.duration(days=2).asSeconds()`?

scripts/src/utils/utils.ts
doodyparizada: you're only using `path` for `fromProjectRoot` which you should remove too

tests/src/utils.spec.ts
```diff
@@ -23,6 +45,32 @@ describe("util functions", () => {
...
+				await throwOnAppEarnLimit(app.id, "total_earn", app.config.limits.minute_total_earn, moment.duration({ minutes: 1 }), 100);
+				expect(true).toBeFalsy(); // should throw and not get here
+			} catch (e) {
+				if (e instanceof MarketplaceError) {
+					const err: MarketplaceError = e;
+					expect(err.code).toBe(4292);
+				} else {
+					throw e;
+				}
+			}
```
doodyparizada: Just as a reminder; we should put this pattern in a helper function

scripts/src/utils/RateLimit.ts
Doc999tor: Don't forget the ttl will be evaluated on every constructor run :)

scripts/src/manage_db_data.ts~HEAD
Doc999tor: I didn't add it, you want me to remove the whole comment?

scripts/src/public/services/orders.ts
Doc999tor: To change it to daily?

PR #222, opened 2018-12-12
rename create script and add option to only update offer's meta image 

liorama:
1. Renamed the create script to db_data as it's purpose is not to create the DB anymore as much as it's to manage DB data (I thought of manage_db_data but opted for the shorter version, I'm willing to change it).
2. Added option to only update earn offer's meta image (thumbnails) - tested locally
3. Add only update existing and don't create new (earn) offers option
4. Added a `.github` file with code owners (this will mostly auto tag us for reviews)

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
 		"build": "npm-run-all clean lint transpile",
 		"start": "node scripts/bin/public/index",
 		"start-internal": "node scripts/bin/internal/index",
-		"create-db": "node scripts/bin/create",
+		"create-db": "node scripts/bin/db_data",
 		"start-admin": "node scripts/bin/admin/index",
 		"restart": "npm-run-all build start",
 		"transpile-tests": "tsc -p tests",
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
 		"build": "npm-run-all clean lint transpile",
 		"start": "node scripts/bin/public/index",
 		"start-internal": "node scripts/bin/internal/index",
-		"create-db": "node scripts/bin/db_data",
+		"manage-db-data": "node scripts/bin/manage_db_data",
 		"start-admin": "node scripts/bin/admin/index",
 		"restart": "npm-run-all build start",
 		"transpile-tests": "tsc -p tests",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -67,7 +67,7 @@
 		"build": "npm-run-all clean lint transpile",
 		"start": "node scripts/bin/public/index",
 		"start-internal": "node scripts/bin/internal/index",
-		"create-db": "node scripts/bin/create",
+		"create-db": "node scripts/bin/db_data",
```
nitzantomer: if you want to rename it and this script does a create job then the new name doesn't reflect it, why not `create_db`?

scripts/src/create_data/offers.ts
```diff
@@ -118,6 +120,10 @@ export async function createEarn(
 		console.log("Updating earn offer %s id %s", offer.name, offer.id, options.dryRun ? "(dry run)" : "");
 		content = await OfferContent.findOne({ offerId: offer.id });
 	} else {
+		if (options.onlyUpdate) {
+			console.log(`Skipping offer creation for offer: ${offerName}`);
```
nitzantomer: what's going on???  didn't we say for like a million times that we have spaces between curly brackets???

scripts/src/create_data/offers.ts
```diff
@@ -130,18 +136,24 @@ export async function createEarn(
...
-			description: orderDescription,
-		}
-	};
-	!options.dryRun && await offer.save();
+	if (options.onlyUpdateMetaImage) {
+		offer.meta.image = image;
+	} else {
+		offer.amount = amount;
+		offer.meta = {
+			title, image, description,
```
nitzantomer: i think that if the object spans multiple line then better give each property it's own like

scripts/src/db_data.ts
```diff
@@ -242,20 +252,20 @@ function initArgsParser(): ScriptConfig {
...
-	const readline = require("readline");
-	const prompt = readline.createInterface(process.stdin, process.stdout);
-	return new Promise(resolve => {
-		prompt.question(message + "\n", (answer: string) => {
-			prompt.close();
-			resolve(answer);
-		});
-	});
-}
+// function confirmPrompt(message: string) {
```
nitzantomer: maybe remove it then?

PR #223, opened 2018-12-13
Http context

doodyparizada:
implement logging and authentication with http context lib.
also using nodemon for hot reloading


scripts/src/logging.ts
```diff
@@ -1,7 +1,6 @@
 import * as winston from "winston";
```
doodyparizada: main changes

scripts/src/middleware.ts
```diff
@@ -1,15 +1,15 @@
 import * as express from "express";
```
doodyparizada: main changes

scripts/src/public/routes/index.ts
```diff
@@ -1,23 +1,21 @@
 import * as express from "express";
```
doodyparizada: main changes

scripts/src/public/auth.ts
```diff
@@ -15,3 +16,20 @@ export async function authenticate(req: express.Request): Promise<db.AuthToken>
 
 	return token;
 }
+
+export const authenticateUser = async function(req: express.Request, res: express.Response, next: express.NextFunction) {
+	const token = await authenticate(req);
+	httpContext.set("userId", token.userId);
```
nitzantomer: maybe store these values in the context only AFTER you found the user?

scripts/src/public/auth.ts
```diff
@@ -15,3 +16,20 @@ export async function authenticate(req: express.Request): Promise<db.AuthToken>
...
+
+	const user = await db.User.findOneById(token.userId);
+
+	if (!user) {
+		// This error now defines an inconsistent state in the DB where a token exists but not user is found
+		// This should never happen as the token.user_id is a foreign key to the users table
+		throw TOSMissingOrOldToken();
+	}
+
+	req.context = { user, token };
```
nitzantomer: why is this needed if they are stored in the `httpContext`?

scripts/src/admin/services.ts
```diff
@@ -6,7 +6,7 @@ import { OpenOrderStatus, Order, OrderContext } from "../models/orders";
 import { IdPrefix, isNothing } from "../utils";
 import * as payment from "../public/services/payment";
 import { BlockchainConfig, getBlockchainConfig } from "../public/services/payment";
-import { getDefaultLogger } from "../logging";
+import { getDefaultLogger as log } from "../logging";
```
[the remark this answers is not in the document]
doodyparizada: I want to add logs in more files and this was just a preparation

scripts/src/public/auth.ts
doodyparizada: This way I can use the values as soon as possible so if for some reason the user doesn't exists I can debug it with these values

scripts/src/public/auth.ts
doodyparizada: They're not stored in the httpContext, but you're right. We could also push them into it

doodyparizada: ok

scripts/src/admin/services.ts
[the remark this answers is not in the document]
doodyparizada: !

PR #228, opened 2018-12-18
add duration to rate limit prefix

doodyparizada:
This way when there are 2 rate limits with different durations but the same prefix they don't collide
Also add rate limit on wallet earn, so a user can't earn too much into a single wallet

### DIFF
--- a/scripts/src/utils/rate_limit.ts
+++ b/scripts/src/utils/rate_limit.ts
@@ -2,40 +2,39 @@ import * as moment from "moment";
 
 import { getRedisClient, RedisAsyncClient } from "../redis";
 import { TooManyRegistrations, TooMuchEarnOrdered } from "../errors";
+import { Application } from "../models/applications";
 
-class RateLimit {
+export class RateLimit {
 	private readonly bucketPrefix: string;
-	private readonly rateLimitValue: number;
 	private readonly redis: RedisAsyncClient;
 	private readonly windowSize: number = 0;
 	private readonly bucketSize: number = 0;
 	private readonly currentTimestampSeconds: number = 0;
-	private readonly ttl: number = moment.duration({ days: 2 }).asSeconds(); // two days in seconds
+	private readonly ttl: number; // two days in seconds
 
 	/**
 	 * @param      {string}  bucketPrefix
-	 * @param      {number}  rateLimitValue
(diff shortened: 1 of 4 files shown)


scripts/src/utils/rate_limit.ts
```diff
@@ -1,45 +1,34 @@
...
 
 class RateLimit {
 	private readonly bucketPrefix: string;
-	private readonly rateLimitValue: number;
 	private readonly redis: RedisAsyncClient;
 	private readonly windowSize: number = 0;
 	private readonly bucketSize: number = 0;
 	private readonly currentTimestampSeconds: number = 0;
-	private readonly ttl: number = moment.duration({ days: 2 }).asSeconds(); // two days in seconds
+	private readonly ttl: number; // two days in seconds
```
nitzantomer: i also prefer that the value assignment be done in the ctor, but maybe move the comment as well?

scripts/src/utils/rate_limit.ts
```diff
@@ -68,32 +57,30 @@ class RateLimit {
...
-export async function throwOnRateLimit(appId: string, type: string, limit: number, duration: moment.Duration) {
-	const rateLimitPrefix: string = "rate_limit";
-	const bucketPrefix: string = `${ rateLimitPrefix }:${ appId }:${ type }:`;
-	const rateLimit: RateLimit = new RateLimit(bucketPrefix, limit, duration);
-	const rateCount = await rateLimit.checkRate();
-	if (rateCount > limit) {
-		throw TooManyRegistrations(`app: ${ appId }, type: ${ type } exceeded the limit: ${ limit } with: ${ rateCount }`);
+// throw error when action should be limited
+async function rateLimit(type: string, duration: moment.Duration, limit: number, error: (msg: string) => MarketplaceError, step: number = 1): Promise<void> {
+	const limiter: RateLimit = new RateLimit(type, duration);
```
nitzantomer: can just be: `const limiter = new ...`

nitzantomer:
in addition to what i've written above, i think we need to rename the throwing functions so that they will have the `assert` prefix, so:

- assertRateLimitAppEarn
- assertRateLimitUserEarn
- assertRateLimitWalletEarn

PR #235, opened 2018-12-24
Nitzan/eco 752/logout

nitzantomer:
added a DELETE `users/me/session` endpoint

### DIFF
--- a/scripts/src/public/auth.ts
+++ b/scripts/src/public/auth.ts
@@ -11,7 +11,7 @@ async function getTokenAndUser(req: express.Request): Promise<[db.AuthToken, db.
 	}
 
 	const token = await db.AuthToken.findOneById(req.token);
-	if (!token) {
+	if (!token || !token.valid) {
 		throw InvalidToken(req.token);
 	}
 
--- a/scripts/src/public/auth.ts
+++ b/scripts/src/public/auth.ts
@@ -11,7 +11,7 @@ async function getTokenAndUser(req: express.Request): Promise<[db.AuthToken, db.
 	}
 
 	const token = await db.AuthToken.findOneById(req.token);
-	if (!token || !token.valid) {
+	if (!token || !token.valid || token.isExpired()) {
 		throw InvalidToken(req.token);
 	}
 
--- a/scripts/src/public/routes/users.ts
+++ b/scripts/src/public/routes/users.ts
@@ -5,6 +5,7 @@ import { Application } from "../../models/applications";
 import { InvalidWalletAddress, NoSuchApp, UnknownSignInType } from "../../errors";
 
 import {
+	logout as logoutService,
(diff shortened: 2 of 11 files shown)


scripts/src/public/auth.ts
```diff
@@ -11,7 +11,7 @@ async function getTokenAndUser(req: express.Request): Promise<[db.AuthToken, db.
 	}
 
 	const token = await db.AuthToken.findOneById(req.token);
-	if (!token) {
+	if (!token || !token.valid) {
```
doodyparizada: great catch. I wonder if clients will start failing or that the refresh mechanism will work @YohayBar ?

scripts/src/public/routes/users.ts
```diff
@@ -127,3 +128,8 @@ export const myUserInfo = async function(req: Request, res: Response) {
 	req.params.user_id = req.context.user!.appUserId;
 	await (userInfo as any)(req as UserInfoRequest, res);
 } as any as RequestHandler;
+
+export const logoutUser = async function(req: Request, res: Response) {
+	await logoutService(req.context.user!, req.context.token!);
+	res.status(200).send();
```
doodyparizada: usually if there is no output we use 204 (No Content)

scripts/src/public/routes/users.ts
```diff
@@ -127,3 +128,8 @@ export const myUserInfo = async function(req: Request, res: Response) {
 	req.params.user_id = req.context.user!.appUserId;
 	await (userInfo as any)(req as UserInfoRequest, res);
 } as any as RequestHandler;
+
+export const logoutUser = async function(req: Request, res: Response) {
+	await logoutService(req.context.user!, req.context.token!);
```
doodyparizada: I was thinking that we should use a different `Request` object for authenticated and non authenticated requests. The authenticated version will have context.user and token as definite fields

scripts/src/public/auth.ts
doodyparizada: Maybe we should also check the expiration here. I thought `valid` checked that too

tests/src/services/users.spec.ts
```diff
@@ -131,4 +131,19 @@ describe("api tests for /users", async () => {
...
+
+		await mock(app)
+			.delete("/v1/users/me/session")
+			.send()
+			.set("Authorization", `Bearer ${ token.id }`)
+			.expect(200);
+
+		token = (await AuthToken.findOne({ userId: user.id }))!;
+		expect(token.valid).toBeFalsy();
+	});
```
doodyparizada: NiCe!

scripts/src/public/routes/users.ts
nitzantomer: that's a good idea, but it's a task/PR of it's own

PR #236, opened 2018-12-24
order history only returns order for current wallet

nitzantomer:
for that i also changed the orders contexts table/model to include the wallet.

### DIFF
--- a/scripts/src/public/services/orders.ts
+++ b/scripts/src/public/services/orders.ts
@@ -136,10 +136,11 @@ async function createOrder(appOffer: AppOffer, user: User, userDeviceId: string,
 		}
 	}, {
 		user,
+		wallet: wallet.address,
 		type: appOffer.offer.type,
 		// TODO if order meta content is a template:
 		// replaceTemplateVars(offer, offer.meta.order_meta.content!)
-		meta: orderMeta,
+		meta: orderMeta
 	});
 	await order.save();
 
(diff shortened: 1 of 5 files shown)


scripts/src/public/services/orders.ts
```diff
@@ -370,23 +375,25 @@ export async function cancelOrder(orderId: string): Promise<void> {
...
+	user: User,
+	deviceId: string,
 	filters: { origin?: db.OrderOrigin; offerId?: string; },
 	limit: number = 25,
 	before?: string,
 	after?: string): Promise<OrderList> {
 
 	// XXX use the cursor input values
 	const status: db.OrderStatusAndNegation = "!opened";
+	const wallet = (await user.getWallets(deviceId)).lastUsed();
```
doodyparizada: shouldn't `getWallets(deviceId)` only return a single wallet?  Why don't you create `getWallet(deviceId)` which is separate from the `getWallets()` function?

tests/src/services/orders.spec.ts
```diff
@@ -88,6 +88,34 @@ describe("test orders", async () => {
...
+		for (let i = firstIteration; i < offers.offers.length; i++) {
+			const offerId = offers.offers[i].id;
+			const openOrder = await createMarketplaceOrder(offerId, user, deviceId);
+			const order = await submitOrder(openOrder.id, user, deviceId, "{}", user.appId);
+			await helpers.completePayment(order.id);
+		}
+
+		const history2 = (await getOrderHistory(user, deviceId, {})).orders.map(order => order.id);
+		expect(history1.length + history2.length).toBe(offers.offers.length);
+		history1.forEach(id => expect(history2).not.toContain(id));
```
doodyparizada: nice!

scripts/src/public/services/orders.ts
nitzantomer: a device might have more than one wallet

PR #239, opened 2018-12-26
Nitzan/eco 553/multiple users devices wallets

nitzantomer:


scripts/src/public/services/users.ts
```diff
@@ -124,15 +123,78 @@ export type UserStats = {
...
+	}
+
+	return {
+		user,
+		token: authToken,
+		profile: await createUserProfile(user)
+	};
+}
+
+async function createUserProfile(user: User): Promise<UserProfile> {
```
doodyparizada: weird name for something that queries the DB and doesn't create anything

doodyparizada (commented):
1. tests
2. instead of Old use V1 prefix
3. If you can move all the V1 stuff to a separate file it would be nice

scripts/src/admin/services.ts
```diff
@@ -337,20 +336,24 @@ async function orderToHtml(order: Order): Promise<string> {
 }
 
 async function userToHtml(user: User): Promise<string> {
+	const accounts = (await user.getWallets()).all().map(wallet => {
+		return `
+		<a href="${ BLOCKCHAIN.horizon_url}/accounts/${ wallet.address }">${ wallet.address }</a>
```
ravitb: You could use DomParser, much more readable for html chunk of code.

nitzantomer: the entire admin is written in this awful way, we need to change all of it and go with react

scripts/src/client.ts
```diff
@@ -61,13 +61,13 @@ export class ClientError extends Error {
 }
 
 export class ClientRequests {
-	public static async create(data: { device_id: string; }, headers?: StringMap) {
-		const res = await axios.post<AuthToken>(MARKETPLACE_BASE + "/v1/users", data, { headers });
-		return new ClientRequests(res.data);
+	public static async create(data: any, headers?: StringMap) {
```
doodyparizada: why `any`?

doodyparizada (commented):
1. Would like to see less use of `any`
2. Would like to reduce duplication of code by implementing v1 as a lightweight layer over v2.

scripts/src/public/services/orders.ts
```diff
@@ -398,23 +461,38 @@ function openOrderDbToApi(order: db.Order, userId: string): OpenOrder {
...
+
+	const data: any = {};
+	let context = order.contextForUser(userId)!;
+	if (context) {
+		Object.assign(data, {
+			offer_type: context.type,
+		}, pick(context.meta, "title", "description", "content", "call_to_action"));
+	} else {
+		context = order.contextForWallet(wallet)!;
+		const app = (await Application.findOneById(context.user.appId))!;  // TODO This is run inside a for loop - better remove this
```
doodyparizada: This is run inside a for loop - better remove this

doodyparizada (approved):
Make this changes in a separate PR

PR #240, opened 2018-12-26
node console v2

liorama:
Should load better in docker container and generally be better.
JS file included cause the build doesn't really build.


doodyparizada:
@liorama merge?

PR #241, opened 2018-12-26
remove offer_id from native metrics

doodyparizada:


scripts/src/metrics.ts
```diff
@@ -32,16 +32,16 @@ export function timeRequest(time: number, method: string, path: string, appId: s
 	statsd.timing("request", time, { method, path, app_id: appId });
 }
 
-export function createOrder(orderType: "marketplace" | "external", offerType: "earn" | "spend", offerId: string, appId: string) {
-	statsd.increment("create_order", 1, { order_type: orderType, offer_type: offerType, offer_id: offerId, app_id: appId });
+export function createOrder(orderOrigin: OrderOrigin, offerType: "earn" | "spend" | "p2p", offerId: string, appId: string) {
```
nitzantomer: maybe create a type `type OfferType = "earn" | "spend" | "p2p";`?

scripts/src/public/services/orders.ts
```diff
@@ -267,9 +267,11 @@ export async function createExternalOrder(jwt: string, user: User): Promise<Open
 
 		await order.save();
 
-		order.contexts.forEach(context => {
-			metrics.createOrder("external", context.type, payload.offer.id, user.appId);
-		});
+		metrics.createOrder(
+			"external",
+			order.isP2P() ? "p2p" : order.contexts[0].type,
```
nitzantomer: maybe put this into a function as it's used at least 3 times

PR #245, opened 2019-01-06
Victor/eco 895/checking overriding users

Doc999tor:

### DIFF
--- a/scripts/src/models/users.ts
+++ b/scripts/src/models/users.ts
@@ -32,17 +32,35 @@ export class User extends CreationDateModel {
 	public walletCount!: number;
 
 	public async save(): Promise<this> {
-		await getManager()
-			.createQueryBuilder()
-			.insert()
-			.into(User)
-			.values([{
-				id: this.id,
-				appId: this.appId,
-				appUserId: this.appUserId,
-				walletAddress: this.walletAddress
-			}])
-			.execute();
+		let errorCount = 0;
+		while (true) {
+			if (errorCount > 3) { break; }
+			console.log("errorCount: ", errorCount);
+
+			try {
+				/**
+				 * tries to insert a new user with generated id
+				 */
+				await getManager()
+					.createQueryBuilder()
+					.insert()
+					.into(User)
+					.values([{
+						id: this.id,
+						appId: this.appId,
+						appUserId: this.appUserId,
+						walletAddress: this.walletAddress,
+						walletCount: this.walletCount, // possibly null
+						createdDate: this.createdDate, // possibly null
+					}])
+					.execute();
+				break; // breaks the while loop in case of success
+			} catch (e) {
+				console.log(e.message);
+				this.id = generateId(IdPrefix.User);
+				errorCount++;
+			}
+		}
 		return this;
 	}
 }
--- a/scripts/src/models/users.ts
+++ b/scripts/src/models/users.ts
@@ -14,9 +14,10 @@ import { CreationDateModel, register as Register, initializer as Initializer } f
 @Entity({ name: "users" })
 @Register
 @Initializer("id", () => generateId(IdPrefix.User))
+@Initializer("walletCount", () => User.DEFAULT_WALLET_COUNT)
 // @Unique(["appId", "appUserId"]) // supported from 0.2.0
 export class User extends CreationDateModel {
-	public static readonly defaultWalletCount = 1;
+	public static readonly DEFAULT_WALLET_COUNT = 1;
 
 	@Column({ name: "app_id" })
 	public appId!: string;
@@ -30,7 +31,7 @@ export class User extends CreationDateModel {
 	@OneToMany(type => OrderContext, context => context.user)
 	public contexts!: OrderContext[];
 
-	@Column({ name: "wallet_count", default: User.defaultWalletCount })
+	@Column({ name: "walletCount" })
 	public walletCount!: number;
 
 	/**
@@ -39,7 +40,7 @@ export class User extends CreationDateModel {
 	 * It generates id and tries to insert it to the table, up to 3 tries, and breaks the loops on success
 	 */
 	public async save(): Promise<this> {
-		if (!this.isNew) { await super.save(); return this; }
+		if (!this.isNew) { return await super.save(); }
 
 		let errorCount = 0;
 		const triesCount = 3;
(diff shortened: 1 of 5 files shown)


scripts/src/models/users.ts
```diff
@@ -23,8 +30,44 @@ export class User extends CreationDateModel {
 	@OneToMany(type => OrderContext, context => context.user)
 	public contexts!: OrderContext[];
 
-	@Column({ name: "wallet_count", default: 1 })
+	@Column({ name: "wallet_count", default: User.defaultWalletCount })
```
nitzantomer: you can add an initializer instead (for the class):
`@Initializer("walletCount", () => User.defaultWalletCount))`
this makes more sense if you take into account my comment about the `insert.values` later on

scripts/src/models/users.ts
```diff
@@ -23,8 +30,44 @@ export class User extends CreationDateModel {
...
+		const triesCount = 3;
+		while (true) {
+			if (errorCount > triesCount) { throw new Error(`user generated with the same id more than ${ triesCount } times or some another error`); }
+
+			try { // tries to insert a new user with generated id
+				await getManager()
+					.createQueryBuilder()
+					.insert()
+					.into(User)
+					.values([{
```
nitzantomer: the problem with this is that we'll need to remember that each field needs to be specified here once added to the class, so it's error-prone.
from the docs/definition files/examples it seems like you can just do:
`.values([this])`
and if you use the `@Initializer` decoration (as my comment above suggested) then you also won't have a null in the `this.walletCount` member.

scripts/src/utils/utils.ts
```diff
@@ -69,14 +71,21 @@ export enum IdPrefix {
 const ID_LENGTH = 20;
 const ID_CHARS = "1234567890abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
 
+/**
```
nitzantomer: yes, this is good to have doc comments, but what format have you chosen here?

scripts/src/utils/utils.ts
```diff
@@ -69,14 +71,21 @@ export enum IdPrefix {
...
+ * @param      {IdPrefix}  prefix  The id prefix
+ * @return     {string}  random string of ID_LENGTH length
+ */
 export function generateId(prefix: IdPrefix | string = IdPrefix.None): string {
-	let id = "";
-
-	while (id.length < ID_LENGTH) {
-		id += ID_CHARS[randomInteger(0, ID_CHARS.length)];
-	}
+	const buffer = Buffer.alloc(ID_LENGTH);
```
nitzantomer: interesting.
i thought that making sure we don't override the user on the db is enough, but changing how we generate ids can also work.
i've never used this method, is it optimal? we generate a lot of random ids...

nitzantomer (changes requested):

scripts/src/utils/utils.ts
```diff
@@ -69,14 +71,21 @@ export enum IdPrefix {
...
+	const randomInts = new Uint8Array(crypto.randomFillSync(buffer)); // not async function for saving existing function interface the same
```
doodyparizada: Why can't we change it to async?

scripts/src/utils/utils.ts
doodyparizada: Does this take a while? Can we just allocate a single Buffer once as a global var?

scripts/src/models/users.ts
```diff
@@ -11,6 +16,8 @@ import { CreationDateModel, register as Register, initializer as Initializer } f
 @Initializer("id", () => generateId(IdPrefix.User))
 // @Unique(["appId", "appUserId"]) // supported from 0.2.0
 export class User extends CreationDateModel {
+	public static readonly defaultWalletCount = 1;
```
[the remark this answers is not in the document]
Doc999tor: 1. I use it below: `walletCount: User.defaultWalletCount, ` r59
I tried to use this.defaultWalletCount but TS didn't understand it
2. I'll renamed it to all caps

scripts/src/models/users.ts
```diff
@@ -23,8 +30,44 @@ export class User extends CreationDateModel {
...
+	@Column({ name: "wallet_count", default: User.defaultWalletCount })
 	public walletCount!: number;
+
+	/**
+	 * Overrided save method
+	 * If this (user) is new, it calls direct insert method instead of built-in upsert TypeORM functionality
+	 * It generates id and tries to insert it to the table, up to 3 tries, and breaks the loops on success
+	 */
+	public async save(): Promise<this> {
+		if (!this.isNew) { await super.save(); return this; }
```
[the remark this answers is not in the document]
Doc999tor: Sure, forgot it

scripts/src/utils/utils.ts
Doc999tor: Doxygen, I like it much

scripts/src/utils/utils.ts
Doc999tor: Sure, it's as optimal as generating ID_LENGTH true random numbers can be, plus it uses threadpool under the hood
`randomInteger` function that is much more used I didn't touch

scripts/src/utils/utils.ts
nitzantomer: oh, well, i prefer working with [tsdoc](https://github.com/Microsoft/tsdoc) as it makes more sense because it's for typescript and works like javadoc (it infers the types)

scripts/src/utils/utils.ts
Doc999tor: We can
It's like allocating new small (ID_LENGTH=20bytes) Int8Array array
We create objects and arrays without bytes counting. But I can use it globally if you want.

scripts/src/models/users.ts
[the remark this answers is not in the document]
nitzantomer: but you use it inside the `User` class, so even if it's private it should work just fine (not using `this`)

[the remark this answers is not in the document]
Doc999tor: We moved `User.DEFAULT_WALLET_COUNT` out of the class to an initializer, so it's not relevant

nitzantomer (approved):

PR #249, opened 2019-01-09
Victor/eco 805/test jwt validation

Doc999tor:
Add test for jwt validation

- [x] linked to a JIRA issue

### DIFF
--- a/scripts/src/internal/jwt.ts
+++ b/scripts/src/internal/jwt.ts
@@ -16,7 +16,7 @@ function getKeyForAlgorithm(alg: string): string {
 	return keyid;
 }
 
-export function sign(subject: string, payload: any, alg?: string) {
+export function sign(subject: string, payload: object, alg?: string) {
 	const keyid = getKeyForAlgorithm(alg || "es256");
 	const signWith = PRIVATE_KEYS[keyid];
 	return jsonwebtoken.sign(payload, signWith.key, {
--- a/scripts/src/internal/services.ts
+++ b/scripts/src/internal/services.ts
@@ -16,7 +16,7 @@ import { AssetUnavailable, BlockchainError, WrongAmount, WrongRecipient, WrongSe
 import { setFailedOrder } from "../public/services/orders";
 import { Application, AppOffer } from "../models/applications";
 
-const BLOCKCHAIN = "stellar-testnet";
+const BLOCKCHAIN = "kin-prod";
 const RS512_APPS = ["test", "smpl"];
 
 export type WalletCreationSuccessData = {
--- a/scripts/src/tests/helpers.ts
+++ b/scripts/src/tests/helpers.ts
@@ -1,10 +1,10 @@
 import { getManager } from "typeorm";
 import * as StellarSdk from "stellar-sdk";
 
-import { generateId } from "../utils/utils";
+import { generateId, readKeysDir, random } from "../utils/utils";
 import { Asset, Offer } from "../models/offers";
 import { User, AuthToken } from "../models/users";
-import { Application, ApplicationConfig } from "../models/applications";
+import { Application, ApplicationConfig, StringMap } from "../models/applications";
 import { LimitConfig } from "../config";
 import { createEarn, createSpend } from "../create_data/offers";
 import { Poll, PageType } from "../public/services/offer_contents";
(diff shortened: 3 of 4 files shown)


scripts/src/internal/jwt.ts
```diff
@@ -16,7 +16,7 @@ function getKeyForAlgorithm(alg: string): string {
 	return keyid;
 }
 
-export function sign(subject: string, payload: any, alg?: string) {
+export function sign(subject: string, payload: object, alg?: string) {
```
doodyparizada: @nitzantomer I made this change - reducing the amount of `any` in the code. Does this make sense?

scripts/src/internal/services.ts
```diff
@@ -16,7 +16,7 @@ import { AssetUnavailable, BlockchainError, WrongAmount, WrongRecipient, WrongSe
 import { setFailedOrder } from "../public/services/orders";
 import { Application, AppOffer } from "../models/applications";
 
-const BLOCKCHAIN = "stellar-testnet";
+const BLOCKCHAIN = "kin-prod";
```
doodyparizada: I also accidentally made this change here too. But I think we should keep it instead of adding a new PR

scripts/src/tests/helpers.ts
```diff
@@ -170,6 +174,22 @@ export async function clearDatabase() {
 	}
 }
 
+const CONFIG = getConfig();
+const PRIVATE_KEYS = readKeysDir(path.join(CONFIG.jwt_keys_dir, "private_keys"));
+const PUBLIC_KEYS = readKeysDir(path.join(CONFIG.jwt_keys_dir, "public_keys"));
```
doodyparizada: I think @nitzantomer would like us to move these consts to the top of the file

doodyparizada (dismissed):
to me it looks great. Maybe @nitzantomer has other comments

scripts/src/tests/helpers.ts
nitzantomer: yeah, it makes more sense to have all variable declarations to the top of the module

scripts/src/tests/helpers.ts
```diff
@@ -1,17 +1,21 @@
...
+import { Application, ApplicationConfig, StringMap } from "../models/applications";
 import { LimitConfig } from "../config";
 import { createEarn, createSpend } from "../create_data/offers";
 import { Poll, PageType } from "../public/services/offer_contents";
 import { CompletedPayment, paymentComplete } from "../internal/services";
 import { ExternalOrder, MarketplaceOrder, Order } from "../models/orders";
 import * as payment from "../public/services/payment";
 import { Event } from "../analytics";
+import { getConfig } from "../internal/config";
+import * as path from "path";
```
nitzantomer: please move all node_modules imports to the top of the file

scripts/src/tests/helpers.ts
```diff
@@ -1,17 +1,21 @@
 import { getManager } from "typeorm";
 import * as StellarSdk from "stellar-sdk";
 
-import { generateId } from "../utils/utils";
+import { generateId, readKeysDir, random } from "../utils/utils";
 import { Asset, Offer } from "../models/offers";
 import { User, AuthToken } from "../models/users";
-import { Application, ApplicationConfig } from "../models/applications";
+import { Application, ApplicationConfig, StringMap } from "../models/applications";
```
nitzantomer: please move the `StringMap` thing to another file, maybe the `utils/utils` module or even create a new `utils/types` module?

scripts/src/tests/services/users.spec.ts
```diff
@@ -125,4 +129,43 @@ describe("api tests for /users", async () => {
...
+	test("testSign", async () => {
+		const app = await createApp(generateId(IdPrefix.App));
+		const payload = { test: "test" };
+		const jwt = await signJwt(app.id, "subject", payload);
+		const res = await verify<{ test: string }, "test_subject">(jwt);
+		expect(res.payload.test).toEqual(payload.test);
+	});
+
+	test("testMalformedJWT", async () => {
+		let jwt;
```
nitzantomer: please annotate these two with types

nitzantomer (changes requested):

scripts/src/tests/helpers.ts
Doc999tor: Done

scripts/src/tests/helpers.ts
Doc999tor: There are many uses of `StringMap`, I can do it in separate PR

scripts/src/tests/services/users.spec.ts
Doc999tor: What type `payload` should be?

nitzantomer: you can go with `object`, as defined in the function itself

nitzantomer (approved):

PR #255, opened 2019-01-15
ECO-974 catch body-parser errors and convert to marketplace errors instead of 5xx

doodyparizada:
Once I removed the nginx proxy, we started receiving:
```
Error
    method: POST
    path: /v1/users
    payload: {}
    message: request aborted
    stack: BadRequestError: request aborted
    at IncomingMessage.onAborted (/opt/app/node_modules/raw-body/index.js:231:10)
    at IncomingMessage.emit (events.js:182:13)
    at abortIncoming (_http_server.js:449:9)
    at socketOnClose (_http_server.js:442:3)
    at Socket.emit (events.js:187:15)
    at TCP._handle.close (net.js:610:12)
```

The error is thrown by express-body-parser and is handled as a 5xx error, when it should be handled as a client error.

### DIFF
--- a/scripts/src/middleware.ts
+++ b/scripts/src/middleware.ts
@@ -11,6 +11,7 @@ import { generateId, getAppIdFromRequest } from "./utils/utils";
 import { MarketplaceError } from "./errors";
 import { getDefaultLogger as logger } from "./logging";
 import { abort as restartServer } from "./server";
+import { HttpError } from "http-errors";
 
 const START_TIME = (new Date()).toISOString();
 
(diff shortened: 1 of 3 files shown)


scripts/src/middleware.ts
```diff
@@ -69,6 +70,11 @@ export const notFoundHandler = function(req: Request, res: Response) {
 export function generalErrorHandler(err: any, req: Request, res: Response, next: express.NextFunction) {
 	if (err instanceof MarketplaceError) {
 		clientErrorHandler(err, req as express.Request, res);
+	} else if (err instanceof HttpError) {
```
liorama: Can't this also catch 5xx errors created with http-errors?

liorama: But what about 5xx, won't they will also be an instance of http-errors? also, what if express changes dependency? (where did you see that errors are of this type?). There must be a better way to test here. Why not for the error status being of 4xx?

liorama (changes requested):

PR #256, opened 2019-01-16
Victor/eco 959/filtering out malformed offers

Doc999tor:
- [x] linked to a JIRA issue

### DIFF
--- a/scripts/src/tests/helpers.ts
+++ b/scripts/src/tests/helpers.ts
@@ -5,7 +5,7 @@ import { getManager } from "typeorm";
 import * as StellarSdk from "stellar-sdk";
 
 import { generateId, readKeysDir, random } from "../utils/utils";
-import { Asset, Offer } from "../models/offers";
+import { Asset, Offer, OfferContent } from "../models/offers";
 import { User, AuthToken } from "../models/users";
 import { Application, ApplicationConfig, StringMap } from "../models/applications";
 import { LimitConfig } from "../config";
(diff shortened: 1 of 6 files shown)


scripts/src/tests/helpers.ts
```diff
@@ -180,6 +182,26 @@ export async function createOffers() {
...
+	const offerContents = (await OfferContent.findOne({ where: { name: offerName } }))!;
+	await offerContents.save(); // TypeORM parses the offers content json and saves it back to the table (it gets in not escaped) - such offer content is malformed and has to be filtered out in every .getAll
+
+	/*	sql for check whether the db has one malformed and wellformed offer content
+		```
+			select * from (select * from offer_contents where content like '{%' limit 1)
+			union all
+			select * from (select * from offer_contents order by id limit 1)
+		```
+	 */
```
doodyparizada: delete this

scripts/src/tests/helpers.ts
```diff
@@ -180,6 +182,26 @@ export async function createOffers() {
 	}
 }
 
+async function createMalformedOffer(num: number) {
+	const offerName = generateId();
+	await createEarn(
+		`${ offerName }_earn${ num }`,
+		"GBOQY4LENMPZGBROR7PE5U3UXMK22OTUBCUISVEQ6XOQ2UDPLELIEC4J",
+		`earn${ num }`, `earn${ num }`, `earn${ num }`, `earn${ num }`, 100, 30, 1, `earn${ num }`, `earn${ num }`, "poll", animalPoll, ["ALL"]
```
doodyparizada: more line breaks

doodyparizada (changes requested):

scripts/src/tests/helpers.ts
Doc999tor: It's malformed offer checking script, I put it here in purpose
Delete it?

scripts/src/tests/helpers.ts
Doc999tor: I copied it from the same function above
I'll add line breaks

doodyparizada: ok

scripts/src/tests/helpers.ts
doodyparizada: Yes

scripts/src/tests/helpers.ts
```diff
@@ -182,24 +181,40 @@ export async function createOffers() {
...
-			select * from (select * from offer_contents order by id limit 1)
-		```
-	 */
+	malformedOfferContent.content = "{a: 1}";
+	await malformedOfferContent.save();
+
+	return offer.id;
+}
+
+export async function createAppUser(offer: Offer, appId: string): Promise<User> {
```
doodyparizada: It's a weird name for what this function does once you removed it from the context. Maybe rename this createAppUserWithOffer

Doc999tor: I just moved existed inner function (closure) from another place
I'll change its name

doodyparizada (approved):

doodyparizada (approved):

PR #257, opened 2019-01-16
ECO-982 More info admin

doodyparizada:
Add userWallet when viewing order history
Add all order attributes (all contexts) when viewing orders (instead of leaving some items blank)


scripts/src/models/orders.ts
```diff
@@ -214,7 +214,7 @@ export const Order = {
 		const query = OrderImpl.createQueryBuilder("ordr")
 			.innerJoinAndSelect("ordr.contexts", "context")
 			.leftJoinAndSelect("context.user", "user")
-			.orderBy("ordr.current_status_date", "DESC")
+			.orderBy("ordr.currentStatusDate", "DESC")
```
doodyparizada: This is weird. It didn't kept failing without changing this but this code is in production too

doodyparizada:
tested on beta

scripts/src/admin/services.ts
```diff
@@ -487,23 +496,19 @@ export async function getApplicationUserData(params: { app_user_id: string, app_
...
-	}
 	if (query.user_id) {
-		const contexts = await OrderContext.find({ userId: query.user_id });
-		if (contexts.length > 0) {
-			q.andWhere("id in (:ids)", { ids: contexts.map(c => c.orderId) });
-		} else {
-			q.andWhere("id = 'no-such-id' ");
+		const userOrders = await Order.genericGet({ userId: query.user_id }).getMany();
+		let orderIds = userOrders.map(o => o.id);
+		if (!orderIds.length) {
```
liorama: 👍

scripts/src/admin/services.ts
```diff
@@ -487,23 +496,19 @@ export async function getApplicationUserData(params: { app_user_id: string, app_
...
+			orderIds = ["no_orders"];
 		}
+		q.where(`ordr.id IN (:ids)`, { ids: orderIds });
+
 	}
-	const orders = await q.orderBy("current_status_date", "DESC").skip(skip(query)).take(take(query)).getMany();
+
+	const orders = await q.skip(skip(query)).take(take(query)).getMany();
```
liorama: Is the skip and take for pagination?

liorama (dismissed):

scripts/src/admin/services.ts
doodyparizada: yes

scripts/src/admin/services.ts
doodyparizada: learnt from the best

PR #260, opened 2019-01-18
Fix rate limit bucket calculation

doodyparizada:
Issue was that rate limiter took the current time in seconds and didn't convert it to the closest bucket.
This cause that any rate limiter over a minute duration, skipped counts allowing more requests in

### DIFF
--- a/scripts/src/utils/rate_limit.ts
+++ b/scripts/src/utils/rate_limit.ts
@@ -2,15 +2,14 @@ import * as moment from "moment";
 
 import { getRedisClient, RedisAsyncClient } from "../redis";
 import { MarketplaceError, TooManyRegistrations, TooMuchEarnOrdered } from "../errors";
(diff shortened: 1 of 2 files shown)


scripts/src/utils/rate_limit.ts
```diff
@@ -2,15 +2,14 @@ import * as moment from "moment";
...
-	private readonly bucketSize: number = 0;
-	private readonly currentTimestampSeconds: number = 0;
-	private readonly ttl: number;
+export class RateLimit {
+	public readonly bucketPrefix: string;
+	public readonly redis: RedisAsyncClient;
+	public readonly windowSize: number = 0;
+	public readonly bucketSize: number = 0;
+	public readonly currentTimestampSeconds: number = 0;
+	public readonly ttl: number;
```
liorama: @doodyparizada why did you make these public?

PR #262, opened 2019-01-21
added current wallet to register response

nitzantomer:


doodyparizada (commented):
This  isn’t exactly what Oren asked for but I think this is the better approach.  He asked that the register call will return the wallet but I think that it should only be queried when needed

PR #264, opened 2019-01-22
Fix duplicate offers

doodyparizada:
There is a unit test that tests this and it passes but apparently the behaviour is different for sqlite and postgres.

I added a test for the duplicate offers in the system tests, made sure it fails and then fixed it.

I also fixed an issue with out expectToThrow logic (which was wrong).

### DIFF
--- a/scripts/src/mock_client.ts
+++ b/scripts/src/mock_client.ts
@@ -426,8 +426,10 @@ async function outdatedJWT() {
 
 	jwt = await appClient.getRegisterJWT(userId, moment().unix(), moment().subtract(1, "days").unix());
 
-	await expectToThrow(() => MarketplaceClient.create({ jwt }),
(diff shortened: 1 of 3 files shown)


scripts/src/mock_client.ts
```diff
@@ -411,18 +421,13 @@ async function outdatedJWT() {
 	const appClient = new SampleAppClient();
 
 	let jwt = await appClient.getRegisterJWT(userId, moment().add(1, "days").unix());
-	try {
-		await MarketplaceClient.create({ jwt });
-		throw new Error("shouldn't be able to register with JWT with iat in the future");
```
doodyparizada: throwing the error here is useless because you catch it the next line... This is fixed in `expectToThrow`

nitzantomer (approved):

PR #273, opened 2019-01-28
only report user_registration_requested when it's a register call

nitzantomer:
if the user is already an existing one there's no need to send this event


doodyparizada:
I prefer it the way it was before

PR #275, opened 2019-01-29
make sure that when paying to self you don't get an earn order

nitzantomer:
we had a bug where p2p to self would result in an external earn order and the user (self) would actually get the kin and the balance would go up.
this makes sure that it won't happen, plus an added test to check it.


doodyparizada (approved):
looks ok

PR #276, opened 2019-01-30
Lior/eco 632 translations tests

liorama:
Tests for translations
- [x] linked to a JIRA issue


scripts/src/manage_db_data.ts
```diff
@@ -321,26 +329,55 @@ initModels(scriptConfig.create_db).then(async () => {
 			const contentType = parsed[0][0].split(/ +/, 2)[1].toLowerCase() as ContentType;
 			let results = [];
 			if (title === "Spend") {
-				results = await parseSpend(parsed, appList);
+				results = await
+					parseSpend(parsed, appList);
```
doodyparizada: @liorama I think you have some different style settings

PR #277, opened 2019-01-30
safer wallet migrate

doodyparizada:

### DIFF
--- a/scripts/src/models/users.ts
+++ b/scripts/src/models/users.ts
@@ -47,12 +47,7 @@ export class User extends CreationDateModel {
 
 		const wallets = await Wallet.find(conditions);
 		if (wallets.length === 0 && this.walletAddress) {
-			deviceId = deviceId || (await AuthToken.findOne({
-				where: { userId: this.id },
-				order: { createdDate: "DESC" }
-			}))!.deviceId;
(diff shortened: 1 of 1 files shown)


scripts/src/models/users.ts
```diff
@@ -47,12 +47,7 @@ export class User extends CreationDateModel {
 
 		const wallets = await Wallet.find(conditions);
 		if (wallets.length === 0 && this.walletAddress) {
-			deviceId = deviceId || (await AuthToken.findOne({
-				where: { userId: this.id },
-				order: { createdDate: "DESC" }
-			}))!.deviceId;
-			logger().info(`lazy migrate user ${ this.id } device ${ deviceId } wallet: ${ this.walletAddress }`);
-			await this.updateWallet(deviceId, this.walletAddress);
+			await this.lazyMigrateWallet(deviceId);
```
YohayBar: @doodyparizada please provide description how this solving Eco-1011 - how can token can be null?

nitzantomer (dismissed):

scripts/src/models/users.ts
doodyparizada: I'm guessing token can be null when lazy migrate is called from either register endpoint with an existing user from a new device. This code doesn't assume we always get a token so it doesn't really matter why the token was null to begin with.

doodyparizada:
Tests do pass it's something buggy with JEST :(

PR #278, opened 2019-01-30
Eco 691 depreacted error

doodyparizada:


scripts/src/errors.ts
```diff
@@ -291,3 +294,11 @@ export function TooManyRegistrations(message: string): MarketplaceError {
 export function TooMuchEarnOrdered(message: string): MarketplaceError {
 	return TooManyRequests(CODES.TooManyRequests.Amounts, message);
 }
+
+function DeprecationError(index: number, message: string): MarketplaceError {
+	return new MarketplaceError(410, index, "Request to a deprecated resource", message);
+}
+
+export function BlockchainEndpointChanged(message: string): MarketplaceError {
+	return TooManyRequests(CODES.Deprecated.BlockchainEndpointChanged, message);
```
YohayBar: TooManyRequests  should be DeprecationError?

scripts/src/errors.ts
```diff
@@ -291,3 +294,11 @@ export function TooManyRegistrations(message: string): MarketplaceError {
 export function TooMuchEarnOrdered(message: string): MarketplaceError {
 	return TooManyRequests(CODES.TooManyRequests.Amounts, message);
 }
+
+function DeprecationError(index: number, message: string): MarketplaceError {
+	return new MarketplaceError(410, index, "Request to a deprecated resource", message);
```
YohayBar: why are we using 410 and not costs?

scripts/src/public/middleware.ts
```diff
@@ -3,13 +3,23 @@ import * as bearerToken from "express-bearer-token";
...
+	next();
+} as express.RequestHandler;
 
 export function init(app: express.Express) {
 	app.use(httpContext.middleware as express.RequestHandler);
 	app.use(bearerToken());
 	app.use(requestLogger);
 	app.use(logRequest);
 	app.use(reportMetrics);
+	app.use(deprecationError);
```
YohayBar: will there be an overhead of checking every request for this specific header?

scripts/src/errors.ts
doodyparizada: `consts`? This is the same throughout the file. I don't mind changing it

scripts/src/errors.ts
doodyparizada: yep. I haven't added a test yet, I opened the PR just for it to be triggered by travis. But this should be caught by a unit test

scripts/src/public/middleware.ts
doodyparizada: It's just another if, and I can also remove it on production for now. The client just needs a way to test the error functionality

scripts/src/errors.ts
```diff
@@ -98,177 +130,177 @@ export class MarketplaceError extends Error {
 }
 
 function UnauthorizedError(index: number, message: string) {
-	return new MarketplaceError(401, index, "Unauthorized Request", message);
+	return new MarketplaceError(CODES.Unauthorized.code, index, "Unauthorized Request", message);
 }
 
 export function MissingToken() {
-	return UnauthorizedError(CODES.Unauthorized.MissingToken, "Request missing token");
+	return UnauthorizedError(CODES.Unauthorized.types.MissingToken, "Request missing token");
```
nitzantomer: this is becoming too verbose, how about:
```
function UnauthorizedError(key: keyof typeof CODES.Unauthorized.types, message: string) {
	return new MarketplaceError(CODES.Unauthorized.code,, CODES.Unauthorized.types[key], "Unauthorized Request", message);
}
```

And then:
```
export function MissingToken() {
    return UnauthorizedError("MissingToken", "Request missing token");
}
```

scripts/src/public/middleware.ts
nitzantomer: this is weird. not that we should care about the extra function for each request, it's just that this is needed for a very specific thing only on dev/test.
why can't we just open a specific endpoint for it?

nitzantomer (approved):

PR #280, opened 2019-01-31
replace the description with the right type

nitzantomer:
the type should be the content type and not offer type
also, added cache to offers, and moved the offer content get to the class

### DIFF
--- a/scripts/src/admin/services.ts
+++ b/scripts/src/admin/services.ts
@@ -7,7 +7,6 @@ import { IdPrefix, isNothing } from "../utils/utils";
 import * as payment from "../public/services/payment";
 import { BlockchainConfig, getBlockchainConfig } from "../public/services/payment";
 import { getOffers as getUserOffersService } from "../public/services/offers";
-import { getOfferContent } from "../public/services/offer_contents";
 
 type OfferStats = {
 	id: string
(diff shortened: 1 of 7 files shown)


scripts/src/admin/services.ts
```diff
@@ -668,7 +667,7 @@ function isInOffer(key: string, offer: Offer): key is keyof Offer {
 }
 
 export async function changeOffer(body: ChangeOfferData, params: { offer_id: string }, query: any): Promise<any> {
-	const offer = await Offer.findOneById(params.offer_id);
+	const offer = await Offer.get(params.offer_id);
```
doodyparizada: Maybe the admin shouldn't use the cache

scripts/src/public/services/orders.ts
```diff
@@ -452,7 +448,12 @@ async function orderDbToApi(order: db.Order, userId: string, wallet: string): Pr
 		}, pick(context.meta, "title", "description", "content", "call_to_action"));
 
 		data.title = app.name;
-		data.description = order.isMarketplaceOrder() ? capitalizeFirstLetter(context.type) : "Completed";
+		if (order.isMarketplaceOrder) {
+			const offerContent = (await offerDb.OfferContent.get(order.offerId))!;
```
doodyparizada: This is good for now, but a better solution would be to include this piece of information inside the order row, so we don't have to query for it on each order. This can be a big performance hit

doodyparizada: @nitzantomer can you add a ticket for this?

doodyparizada (dismissed):
Looks good. I asked you to create a ticket for something I noticed.

scripts/src/public/services/orders.ts
nitzantomer: @doodyparizada 
https://kinecosystem.atlassian.net/browse/ECO-1034
it's on you

doodyparizada (approved):

PR #282, opened 2019-02-04
authenticated requests

nitzantomer:
now the request objects for authenticated request contain both token and auth context.

### DIFF
--- a/scripts/src/public/auth.ts
+++ b/scripts/src/public/auth.ts
@@ -1,21 +1,42 @@
 import * as express from "express";
-
-import * as db from "../models/users";
-import { MissingToken, InvalidToken, TOSMissingOrOldToken } from "../errors";
 import * as httpContext from "express-http-context";
-import { AuthToken, User } from "../models/users";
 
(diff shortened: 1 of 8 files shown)


scripts/src/public/auth.ts
```diff
@@ -1,21 +1,42 @@
...
+
+export type TokenedRequest = express.Request & {
+	readonly token: string;
+};
+
+export type AuthenticatedRequest = TokenedRequest & {
+	readonly context: AuthContext;
+};
+
+export function isTokenedRequest(req: express.Request): req is TokenedRequest {
```
doodyparizada: remove export

scripts/src/tests/services/users.v1.spec.ts
```diff
@@ -52,18 +52,24 @@ describe("api tests for v1 users", async () => {
 		const user2 = await helpers.createUser({ appId });
 		const token: AuthToken = (await AuthToken.findOne({ userId: user1.id }))!;
 
+		console.log("moo 1");
```
doodyparizada: remove

doodyparizada (dismissed):

doodyparizada (approved):

PR #283, opened 2019-02-05
event user_logout_server_requested was not sent

nitzantomer:


doodyparizada (approved):
Maybe we can cause the constructor to call report. itself so things like this won't happen again

PR #284, opened 2019-02-06
fixed the race condition between logout and login

nitzantomer:
also added the logout and login functionality to the testing client


scripts/src/mock_client.ts
```diff
@@ -1543,7 +1567,9 @@ async function main() {
 	await v1P2p();
```
doodyparizada: uncomment all the tests.
Maybe we can convert this into jest and choose to run a single function from command line

doodyparizada (changes requested):
bring back the commented tests

PR #285, opened 2019-02-06
ECO-1017 count unique wallets when counting for wallet limit

liorama:
Problem was that for the user wallet limit we counted wallets+device as unique when we should count only wallet address. That caused false positives which meant users that use the backup restore feature will be limited to less wallets then the actual limit.

- [x] linked to a JIRA issue


scripts/src/models/users.ts
```diff
@@ -181,6 +181,10 @@ export class Wallets {
 		return this.items.length;
 	}
 
+	public get uniqueCount() {
```
nitzantomer: why do we need both?
why not just change the `count` getter to return the unique count?

liorama: a. `count` sounds like a method we might use.
but mostly b. Not sure it's not used already in other places (amd wasn't too bothered to search for it.)

liorama: @nitzantomer Since it's pretty generic and straightforward implementation I would leave it be but if you want me to remove it and sure it's not used elsewhere, I will. Just let me know, I want to merge this.

liorama: It is used in ```Wallets.lastUsed()``` so I'm leaving it as it is. Can you approve this PR now?

nitzantomer: @liorama 
but the idea behind `count` was to return a unique count, it was a bug that it didn't.
the fact that it's being used is irrelevant, as it uses a bugged function.
there's no reason to have to different getters, especially since one of them doesn't function as it should have.

scripts/src/tests/services/orders.spec.ts
```diff
@@ -653,7 +653,7 @@ describe("test v2 orders", async () => {
 		const orders2 = await createOrdersForUser(user, deviceId2, app, { divideBy: 2 });
 		expect(orders2.length).toBeGreaterThan(0);
 
-		expect((await user.getWallets()).count).toBe(2);
+		expect((await user.getWallets()).count).toBe(1);
```
liorama: @nitzantomer FYI had to change your test

scripts/src/models/users.ts
liorama: Ok got you, done, I thought it was the intended behaviour for other uses.

liorama: @nitzantomer

scripts/src/tests/services/orders.spec.ts
nitzantomer: makes sense!
good.

nitzantomer (approved):

PR #292, opened 2019-02-21
ECO-1095 deny update with wallet from different app

liorama:
Deny restore of wallet if of different app.

- [x] linked to a JIRA issue


scripts/src/public/services/users.ts
```diff
@@ -92,23 +92,38 @@ export type UpdateUserProps = {
...
+		const appWallet = await WalletApplication.findOne({ walletAddress });
+		if (!appWallet) {
+			const newWallet = WalletApplication.create({ walletAddress, appId });
+			await newWallet.save();
+		} else {
+			if (appWallet.appId !== appId) {
+				throw CrossAppWallet();
+			}
+		}
+		/*  End of Cross-app restore check */
```
doodyparizada: instead of putting comments around this section, you can extract it into a function

scripts/src/tests/services/users.spec.ts
```diff
@@ -321,4 +321,36 @@ describe("api tests for v2 users", async () => {
...
+		const token1 = (await AuthToken.findOne({ userId: user1.id }))!;
+		const token2 = (await AuthToken.findOne({ userId: user2.id }))!;
+		await mock(expressApp)
+			.patch("/v2/users/me")
+			.send({ wallet_address: (await user1.getWallets()).first!.address })
+			.set("content-type", "application/json")
+			.set("Authorization", `Bearer ${token1.id}`);
+
+		// console.log(await WalletApplication.find());
+		// console.log("user2", (await user2.getWallets()).first!.address);
```
doodyparizada: remove these comments

scripts/src/public/services/users.ts
```diff
@@ -92,23 +92,38 @@ export type UpdateUserProps = {
...
+			metrics.maxWalletsExceeded(appId);
 			throw MaxWalletsExceeded();
 		}
 
-		const isNewWallet = await user.updateWallet(props.deviceId, props.walletAddress);
+		const isNewWallet = await user.updateWallet(props.deviceId, walletAddress);
 		if (isNewWallet) {
-			logger().info(`creating stellar wallet for user ${ user.id }: ${ props.walletAddress }`);
-			await payment.createWallet(props.walletAddress, user.appId, user.id);
+			logger().info(`creating stellar wallet for user ${appId}: ${props.walletAddress}`);
```
nitzantomer: maybe it's time to fix your ide definitions to have spaces in string templates?

scripts/src/public/services/users.ts
```diff
@@ -121,7 +136,7 @@ export async function activateUser(authToken: DbAuthToken, user: User): Promise<
 
 export async function userExists(appId: string, appUserId: string): Promise<boolean> {
 	const user = await User.findOne({ appId, appUserId });
-	logger().debug(`userExists service appId: ${ appId }, appUserId: ${ appUserId }, user: `, user);
+	logger().debug(`userExists service appId: ${appId}, appUserId: ${appUserId}, user: `, user);
```
nitzantomer: here too

scripts/src/tests/helpers.ts
```diff
@@ -44,7 +44,7 @@ export async function createUser(options: { appId?: string; deviceId?: string; c
 
 	const user = await (User.new(Object.assign(userData, { isNew: true }))).save();
 	if (options.createWallet === undefined || options.createWallet) {
-		await user.updateWallet(deviceId, `test_wallet_${ uniqueId }`);
+		await user.updateWallet(deviceId, `test_wallet_123456789012345678901234${ uniqueId }`);
```
nitzantomer: not that i mind, but why was this change necessary?

scripts/src/tests/services/users.spec.ts
```diff
@@ -112,27 +112,27 @@ describe("api tests for v2 users", async () => {
...
 			.set("x-request-id", "123")
-			.set("Authorization", `Bearer ${ token.id }`)
+			.set("Authorization", `Bearer ${token.id}`)
 			.expect(404);
 	});
 
 	test("updateUser", async () => {
 		const testApp = await helpers.createApp(generateId(IdPrefix.App));
-		const newWalletAddress = "new_address_must_be_56_characters____bla___bla___bla____";
+		const newWalletAddress = `test_wallet_123456789012345678901234${ generateId() }`;
```
nitzantomer: i think that this misses the point.
your change makes the wallet address to have less than 56 chars.

and also, why changing it to begin with?

scripts/src/tests/services/users.spec.ts
```diff
@@ -166,10 +166,10 @@ describe("api tests for v2 users", async () => {
 		let token = (await AuthToken.findOne({ userId: user.id }))!;
 		expect(token.valid).toBeTruthy();
 
-		await mock(app)
+		await mock(expressApp)
 			.delete("/v2/users/me/session")
 			.send()
-			.set("Authorization", `Bearer ${ token.id }`)
+			.set("Authorization", `Bearer ${token.id}`)
```
nitzantomer: here too

scripts/src/tests/services/users.spec.ts
```diff
@@ -321,4 +321,36 @@ describe("api tests for v2 users", async () => {
...
+
+		const token1 = (await AuthToken.findOne({ userId: user1.id }))!;
+		const token2 = (await AuthToken.findOne({ userId: user2.id }))!;
+		await mock(expressApp)
+			.patch("/v2/users/me")
+			.send({ wallet_address: (await user1.getWallets()).first!.address })
+			.set("content-type", "application/json")
+			.set("Authorization", `Bearer ${token1.id}`);
+
+		// console.log(await WalletApplication.find());
```
nitzantomer: remove?

scripts/src/tests/services/users.spec.ts
```diff
@@ -321,4 +321,36 @@ describe("api tests for v2 users", async () => {
...
+			.set("content-type", "application/json")
+			.set("Authorization", `Bearer ${token2.id}`);
+
+		await mock(expressApp)
+			.patch("/v2/users/me")
+			.send({ wallet_address: (await user2.getWallets()).first!.address })
+			.set("content-type", "application/json")
+			.set("Authorization", `Bearer ${token1.id}`)
+			.expect(401);
+		// console.log(await WalletApplication.find());
```
nitzantomer: remove?

nitzantomer (changes requested):

scripts/src/tests/services/users.spec.ts
liorama: It would not pass the new check this PR brings in.

liorama: and no, it's not less than 56 chars hence why the tests pass

scripts/src/tests/helpers.ts
liorama: less than 56 chars, don't know how this passed tests before

scripts/src/public/services/users.ts
liorama: I would still leave the comments around the function call, just to make it clearer so I don't see any benefit. If you really want I'll do it.

nitzantomer: i also don't like it too much.
the first comment is ok, but the end part is just weird.

scripts/src/tests/services/users.spec.ts
nitzantomer: well, you count on `generateId` to return a specific amount of chars, which isn't wise.
we might change that someday, and that will mess things up here.

scripts/src/public/services/users.ts
liorama: It's just to confine the code associated with this so it's clear what needs to be delete when we eventually need to remove it in a few months.

liorama: And also that development until then won't change/intermingled with this code by mistake.

scripts/src/tests/services/users.spec.ts
liorama: mmm I agree, but since there are other places in the tests I've seen it used in a similar manner (at least one instance is above) I don't think it makes any difference now. Also, I think if we do change the length it might make us check production code if our tests break. Although a specific test for this would be better.

scripts/src/tests/services/users.spec.ts
```diff
@@ -121,14 +121,14 @@ describe("api tests for v2 users", async () => {
 
 	test("updateUser", async () => {
 		const testApp = await helpers.createApp(generateId(IdPrefix.App));
-		const newWalletAddress = "new_address_must_be_56_characters____bla___bla___bla____";
+		const newWalletAddress = generateRandomString({ length: 56 });
```
liorama: @nitzantomer happy? (That was a complete waste of my time. thank you). Now please approve and merge it if you're already at it.

nitzantomer: @liorama, no one told you to add functionality which isn't needed. you could have just left it with the old value which had 56 chars...

scripts/src/utils/utils.ts
```diff
@@ -72,12 +72,30 @@ const ID_CHARS = "1234567890abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ
...
  */
 export function generateId(prefix: IdPrefix | string = IdPrefix.None): string {
-	const buffer = Buffer.alloc(ID_LENGTH);
-	const randomInts = new Uint8Array(crypto.randomFillSync(buffer)); // not async function for saving existing function interface the same
+	return generateRandomString({ prefix, minLength: ID_LENGTH });
+}
 
-	return prefix + randomInts.reduce(
-		(str, int) => str + ID_CHARS[Math.trunc(int / 256 * ID_CHARS.length)], ""
+type GenerateRandomStringOptions = {
```
nitzantomer: this should be exported

nitzantomer (dismissed):

scripts/src/tests/services/users.spec.ts
liorama: @nitzantomer No, that would have failed the new test, it needed to be a new address every time.

nitzantomer (dismissed):

PR #293, opened 2019-02-24
Victor/eco 1097/admin config validating

Doc999tor:
- [x] linked to a JIRA issue

### DIFF
--- a/scripts/src/admin/services.ts
+++ b/scripts/src/admin/services.ts
@@ -693,13 +693,12 @@ export async function changeOffer(body: ChangeOfferData, params: { offer_id: str
 	await offer.save();
 }
 
-type UpdateAppConfigRequest = Request & {
-	body: ApplicationConfig
-};
+type AppConfig = ApplicationConfig & { [limits: string]: number };
+type UpdateAppConfigRequest = Request & { body: AppConfig; };
 
 export const updateAppConfig = async function(req: UpdateAppConfigRequest, res: Response) {
-	const config = req.body;
-	const isLimitsNumbers = (obj: any, limitName: string) => typeof obj.limits[limitName] === "number";
+	const config: AppConfig = req.body;
+	const isLimitsNumbers = (configObj: AppConfig, limitName: string) => typeof configObj.limits[limitName as keyof AppConfig["limits"]] === "number";
 	if (!config.limits || !Object.keys(config.limits).every(isLimitsNumbers.bind(null, config))) {
 		res.status(400).send("Config data is incorrect");
 		return false;
--- a/scripts/src/admin/services.ts
+++ b/scripts/src/admin/services.ts
@@ -701,7 +701,7 @@ export const updateAppConfig = async function(req: UpdateAppConfigRequest, res: 
 	const config = req.body;
 	if (
 		!config.limits
-		|| !Object.keys(config.limits).every( // keys of config.limits have to be numbers
+		|| !Object.keys(config.limits).every( // values of config.limits have to be numbers
 			limitName => typeof config.limits[limitName] === "number"
 		)
 	) {
--- a/scripts/src/admin/services.ts
+++ b/scripts/src/admin/services.ts
@@ -699,12 +699,8 @@ type UpdateAppConfigRequest = Request & {
 
 export const updateAppConfig = async function(req: UpdateAppConfigRequest, res: Response) {
 	const config = req.body;
-	if (
-		!config.limits
-		|| !Object.keys(config.limits).every( // values of config.limits have to be numbers
-			limitName => typeof config.limits[limitName] === "number"
-		)
-	) {
+	const isLimitsNumbers = (obj, limitName) => typeof obj.limits[limitName] === "number";
+	if (!config.limits || !Object.keys(config.limits).every(isLimitsNumbers.bind(null, config))) {
 		res.status(400).send("Config data is incorrect");
 		return false;
 	}
--- a/scripts/src/admin/services.ts
+++ b/scripts/src/admin/services.ts
@@ -700,7 +700,7 @@ export const updateAppConfig = async function(req: UpdateAppConfigRequest, res: 
 	const config: AppConfig = req.body;
 	const isLimitsNumbers = (configObj: AppConfig, limitName: string) => typeof configObj.limits[limitName as keyof AppConfig["limits"]] === "number";
 	if (!config.limits || !Object.keys(config.limits).every(isLimitsNumbers.bind(null, config))) {
-		res.status(400).send("Config data is incorrect");
+		res.status(400).send("Config data is invalid");
 		return false;
 	}
 
--- a/scripts/src/admin/services.ts
+++ b/scripts/src/admin/services.ts
@@ -698,13 +698,23 @@ type UpdateAppConfigRequest = Request & {
 };
 
 export const updateAppConfig = async function(req: UpdateAppConfigRequest, res: Response) {
+	const config = req.body;
+	if (
+		!config.limits
+		|| !Object.keys(config.limits).every( // keys of config.limits have to be numbers
+			limitName => typeof config.limits[limitName] === "number"
+		)
+	) {
+		res.status(400).send("Config data is incorrect");
+		return false;
+	}
+
 	const app = (await Application.findOne({ id: req.params.application_id }))!;
 	try {
-		app.config = req.body;
+		app.config = config;
 		await app.save();
 	} catch (e) {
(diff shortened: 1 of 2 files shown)


scripts/src/admin/services.ts
```diff
@@ -698,13 +698,23 @@ type UpdateAppConfigRequest = Request & {
 };
 
 export const updateAppConfig = async function(req: UpdateAppConfigRequest, res: Response) {
+	const config = req.body;
+	if (
```
nitzantomer: i'm not a fan of this formatting at all, why not:
```
const predicate = (limitName: string) => typeof config.limits[limitName] === "number";
if (!config.limits || !Object.keys(config.limits).every(predicate)) {
    ...
```

also, maybe check that the `limitName` values match those we have, and that all of them have been passed?

scripts/src/admin/wrapper.html
```diff
@@ -333,15 +333,27 @@ <h1><a href="/">Marketplace Admin</a></h1>
 				const appId = editor.frame.parentNode.dataset.appId;
 				try {
 					json = editor.get();
+					if (
```
nitzantomer: again, this format isn't fun to read, extract the function into a variable, then use it and do it as a one liner.
and as it looks like the same function, just save it in a place where both can use it

scripts/src/admin/wrapper.html
```diff
@@ -333,15 +333,27 @@ <h1><a href="/">Marketplace Admin</a></h1>
...
 						body: JSON.stringify(json),
 						headers: new Headers({'Content-Type': 'application/json'})
 					})
 					if (response.ok) {
 						closeEditor(editor)
+					} else {
+						toast('config is incorrect')
 					}
-				} catch (e) { console.log('invalid json'); }
+				} catch (e) { toast('smth is wrong'); }
```
nitzantomer: we don't do a one line if/else

nitzantomer (changes requested):

scripts/src/admin/services.ts
Doc999tor: Done
"maybe check that the `limitName` values match those we have" - what does it mean?

scripts/src/admin/wrapper.html
Doc999tor: Done

nitzantomer (dismissed):

scripts/src/admin/services.ts
```diff
@@ -699,7 +699,7 @@ type UpdateAppConfigRequest = Request & {
 
 export const updateAppConfig = async function(req: UpdateAppConfigRequest, res: Response) {
 	const config = req.body;
-	const isLimitsNumbers = (obj, limitName) => typeof obj.limits[limitName] === "number";
+	const isLimitsNumbers = (obj: any, limitName: string) => typeof obj.limits[limitName] === "number";
```
nitzantomer: we try to avoid using `any`, only doing it when there's no other way (or the other way is too complicated), which is not the case here.
it should be `obj: LimitConfig`.

also, you shouldn't be able to bind on an arrow function, and there's also no reason to, you can just do:
```
const isLimitsNumbers = (limitName: string) => typeof config.limits[limitName] === "number";
```

as `config` is part of the closure.

nitzantomer (changes requested):

scripts/src/admin/services.ts
Doc999tor: You're right about the `LimitConfig`, forgot the type already exists
About the binding: predicate with `config` is tight coupled and not functional. By patching the arguments and passing `LimitConfig` argument we can move the arrow function out of that scope

nitzantomer: If you move it outside then it make sense, but then you cannot use an arrow function

Doc999tor: Why not? I missed the point
Export this function to utils and use more than once with the patch of configObj

nitzantomer: it's just weird to use bind on an arrow function.  you cannot change the `this` object using bind on arrow functions, which is why it's weird.
i get that you use bind here to save the reference to the argument, but it's just weird to see it.

scripts/src/admin/services.ts
```diff
@@ -693,18 +693,23 @@ export async function changeOffer(body: ChangeOfferData, params: { offer_id: str
...
-	body: ApplicationConfig
-};
+type AppConfig = ApplicationConfig & { [limits: string]: number };
+type UpdateAppConfigRequest = Request & { body: AppConfig; };
 
 export const updateAppConfig = async function(req: UpdateAppConfigRequest, res: Response) {
+	const config: AppConfig = req.body;
+	const isLimitsNumbers = (configObj: AppConfig, limitName: string) => typeof configObj.limits[limitName as keyof AppConfig["limits"]] === "number";
+	if (!config.limits || !Object.keys(config.limits).every(isLimitsNumbers.bind(null, config))) {
+		res.status(400).send("Config data is incorrect");
```
liorama: I'd change `incorrect` to `invalid`. Incorrect deals with wrong numbers, invalid relates to format. This might sound like mere semantics but think about the user, you want Tom to know it's not that he entered bad numbers, you want to tell him "something is wrong with the format of what you've submitted, have another look or go bother Lior about it"

scripts/src/admin/services.ts
```diff
@@ -693,18 +693,23 @@ export async function changeOffer(body: ChangeOfferData, params: { offer_id: str
...
-type UpdateAppConfigRequest = Request & {
-	body: ApplicationConfig
-};
+type AppConfig = ApplicationConfig & { [limits: string]: number };
+type UpdateAppConfigRequest = Request & { body: AppConfig; };
 
 export const updateAppConfig = async function(req: UpdateAppConfigRequest, res: Response) {
+	const config: AppConfig = req.body;
+	const isLimitsNumbers = (configObj: AppConfig, limitName: string) => typeof configObj.limits[limitName as keyof AppConfig["limits"]] === "number";
+	if (!config.limits || !Object.keys(config.limits).every(isLimitsNumbers.bind(null, config))) {
```
liorama: Personally, I'd put `isLimitsNumbers.bind(null, config)` into a const named something like `isConfigLimitsValid` which makes this a little bit more readable and descriptive (and in the future will hold other tests we might want to do on these user inputs. But that's my _personal_ preference.

scripts/src/admin/services.ts
```diff
@@ -693,18 +693,23 @@ export async function changeOffer(body: ChangeOfferData, params: { offer_id: str
...
 }
 
-type UpdateAppConfigRequest = Request & {
-	body: ApplicationConfig
-};
+type AppConfig = ApplicationConfig & { [limits: string]: number };
+type UpdateAppConfigRequest = Request & { body: AppConfig; };
 
 export const updateAppConfig = async function(req: UpdateAppConfigRequest, res: Response) {
+	const config: AppConfig = req.body;
```
liorama: Nice. Glad to see there's someone else not too lazy to do this.

scripts/src/admin/wrapper.html
```diff
@@ -333,15 +333,25 @@ <h1><a href="/">Marketplace Admin</a></h1>
 				const appId = editor.frame.parentNode.dataset.appId;
 				try {
 					json = editor.get();
+					const isLimitsNumbers = (obj, limitName) => typeof obj.limits[limitName] === 'number';
+					if (!json.limits || !Object.keys(json.limits).every(isLimitsNumbers.bind(null, json))) {
+						toast('config limits got incorrect format')
```
liorama: `app limits are of an incorrect format` or `invalid app limits format`

scripts/src/admin/wrapper.html
```diff
@@ -333,15 +333,25 @@ <h1><a href="/">Marketplace Admin</a></h1>
...
+
 					const response = await fetch(`/applications/${appId}/config`, {
 						method: 'PUT',
 						body: JSON.stringify(json),
 						headers: new Headers({'Content-Type': 'application/json'})
 					})
 					if (response.ok) {
 						closeEditor(editor)
+					} else {
+						toast('config is incorrect')
```
liorama: You do:
`config is incorrect (rejected by server)` which will help us debug this in the future.

scripts/src/admin/services.ts
nitzantomer: @liorama 
it's not about being lazy or not, i think it's redundant.

also, here, there's no need for the type annotation.

nitzantomer (approved):

PR #294, opened 2019-02-25
added the error codes as comments in the errors.ts file

nitzantomer:
because sometimes you want to search by an error code and you don't remember that 409 is conflict...

### DIFF
--- a/scripts/src/errors.ts
+++ b/scripts/src/errors.ts
@@ -13,7 +13,7 @@ export type HeaderValue = number | string | string[];
  */
 const CODES = {
 	BadRequest: {
-		code: HttpCodes.BAD_REQUEST,
+		code: HttpCodes.BAD_REQUEST, // 400
 		types: {
 			UnknownSignInType: 1,
 			WrongJwtAlgorithm: 2,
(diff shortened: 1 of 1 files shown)


liorama (approved):
@nitzantomer Nice.
Do you think maybe changing the name to let say `404_NOT_FOUND` would be better (so it's clear any place it appears in the code base)?

I restarted the travis built since it failed and I couldn't see a good reason why that'll happen.

nitzantomer:
@liorama 
but it says that before the comment:
> NOT_FOUND, // 404
isn't that what you want? or am i missing something?

liorama:
@nitzantomer I meant to change the key name, so instead of:
`CODES.Conflict` we'll do `CODES.409_conflict`, so when looking at:
`new MarketplaceError(CODES.Conflict.code, CODES.Conflict.types[key], "Conflict", message)`
You'd know what the code is without following into the implementation. Also, a descriptive name is better than a comment.

nitzantomer:
@liorama 
oh, there's no need for that, imo.
the use of the `MarketplaceError` constructor is internal to this file.
i added the codes because you sometimes need to find out what's the error with code `4095` (for example) because it's part of an error the clients received.
it was harder for find out if the codes weren't in the code

PR #300, opened 2019-03-06
singleton redis client

Doc999tor:
- [x] linked to a JIRA issue


doodyparizada:
Looks good. Let's run this on beta for a while and look for errors, maybe we'll need to handle reconnections now.
I'd also like to see redis connections graph from beta before and after this change

PR #301, opened 2019-03-18
Order wallets by last used in admin

doodyparizada:
+ use travis env vars instead of generating KIN address

### DIFF
--- a/scripts/src/admin/services.ts
+++ b/scripts/src/admin/services.ts
@@ -351,7 +351,9 @@ async function orderToHtml(order: Order): Promise<string> {
 }
 
 async function userToHtml(user: User): Promise<string> {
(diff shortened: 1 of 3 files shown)


scripts/src/admin/services.ts
```diff
@@ -351,7 +351,9 @@ async function orderToHtml(order: Order): Promise<string> {
 }
 
 async function userToHtml(user: User): Promise<string> {
-	const accounts = (await user.getWallets()).all().map(wallet => {
+	const accounts = (await user.getWallets()).all()
+		.sort((w1, w2) => w1.lastUsedDate.valueOf() - w2.lastUsedDate.valueOf())
```
liorama: Do you really need to use `valueOf()` here?

PR #304, opened 2019-03-18
Victor/eco 1166/get blockchain version from http headers

doodyparizada:


scripts/src/errors.ts
```diff
@@ -328,6 +328,6 @@ function DeprecationError(key: keyof typeof CODES.Gone.types, message: string):
 	return new MarketplaceError(CODES.Gone.code, CODES.Gone.types[key], "Request to a deprecated resource", message);
 }
 
-export function BlockchainEndpointChanged(message: string): MarketplaceError {
-	return DeprecationError("BlockchainEndpointChanged", message);
+export function WrongBlockchainVersion(message: string): MarketplaceError {
+	return DeprecationError("WrongBlockchainVersion", message);
```
doodyparizada: CC @nitzantomer, don't think this affects the client, but the error name changed.

scripts/src/internal/services.ts
```diff
@@ -138,9 +138,7 @@ export async function paymentComplete(payment: CompletedPayment) {
 		return;
 	}
 
-	const blockchain_version = order.blockchainData.blockchain_version;
 	order.blockchainData = pick(payment, "transaction_id", "sender_address", "recipient_address");
-	order.blockchainData.blockchain_version = blockchain_version;
```
doodyparizada: 👍

scripts/src/public/app.ts
```diff
@@ -36,6 +36,7 @@ createV1Routes(app, "/v1");
 app.use(notFoundHandler);
 // catch errors
 app.use(generalErrorHandler);
+app.use(clientMigrationCheck);
```
doodyparizada: I think this should be part of `initCustomMiddleware`

scripts/src/public/middleware.ts
```diff
@@ -3,15 +3,29 @@ import * as bearerToken from "express-bearer-token";
...
-const deprecationError = function(req: express.Request, res: express.Response, next: express.NextFunction) {
-	const shouldRaiseError = req.header("x-simulate-deprecation-error");
-	if (shouldRaiseError) {
-		throw BlockchainEndpointChanged("simulated deprecation");
+export const clientMigrationCheck = async function(req: AuthenticatedRequest, res: express.Response, next: express.NextFunction) {
+	const CLIENT_BLOCKCHAIN_HEADER = "x-kin-blockchain-version";
+	const blockchainVersionHeader = req.header(CLIENT_BLOCKCHAIN_HEADER);
+
+	const app = await Application.get(req.context.user.appId);
+	if (!app) { // cached per instance
```
doodyparizada: redundant comment is redundant

scripts/src/public/middleware.ts
```diff
@@ -3,15 +3,29 @@ import * as bearerToken from "express-bearer-token";
...
+		throw NoSuchApp(req.context.user.appId);
 	}
+
+	const isAppMigrated = app.config.blockchain_version === "3";
+	const isAppVersionEqualsToClient = app.config.blockchain_version !== blockchainVersionHeader;
```
doodyparizada: shouldn't you check for equality? i.e. `===`?

scripts/src/public/middleware.ts
doodyparizada: Also I think you might not always have `req.context` - i.e. this might not be an `AuthenticatedRequest`. You should check that first

scripts/src/public/routes/config.ts
```diff
@@ -53,27 +53,3 @@ export const getAppBlockchainVersion = async function(req: GetAppBlockchainVersi
...
-// 			const app_id = req.params.app_id;
-// 			await setAppBlockchainVersionService(app_id, req.body.blockchain_version);
-// 			res.status(200).send();
-// 		} else {
-// 			res.status(401).send();
-// 		}
-// 	} else {
-// 		res.status(403).send();
-// 	}
-// } as any as RequestHandler;
```
doodyparizada: 👍

scripts/src/public/routes/orders.ts
```diff
@@ -116,8 +118,12 @@ export const submitWhitelistOrder = async function(req: WhitelistTransactionRequ
 	logger().info("submit whitelist order", { userId: req.context.user.id, orderId: req.params.order_id });
 
 	const order = await getOrderService(req.params.order_id, req.context.user);
+	const app = await Application.get(req.context.user.appId);
+	if (!app) {
+		throw NoSuchApp(req.context.user.appId);
+	}
 
-	await axios.default.post(getPaymentServiceUrl(order.blockchain_data.blockchain_version!), order);
+	await axios.default.post(getPaymentServiceUrl(app.config.blockchain_version), order);
```
doodyparizada: Remove this `submitWhitelistOrder` function

scripts/src/tests/services/users.spec.ts
```diff
@@ -213,26 +213,26 @@ describe("api tests for v2 users", async () => {
...
+
+	// 	res = await mock(app)
+	// 		.get(`/v2/users/me`)
+	// 		.set("Authorization", `Bearer ${ token.id }`)
+	// 		.set("x-request-id", "123")
+	// 		.set("x-simulate-deprecation-error", "true");
+
+	// 	expect(res.status).toBe(410);
+	// 	expect(res.body.code).toBe(4101);
+	// });
```
doodyparizada: Instead of removing this test, you can change it to test your new x-kin-blockchain-version header

scripts/src/errors.ts
Doc999tor: Yep, because the old name doesn't fit anymore

scripts/src/tests/services/users.spec.ts
Doc999tor: Fixed the test, but `helpers.createApp` hardcoded to kin2
We will add it in future PRs?

PR #305, opened 2019-03-24
Eco 1047/tests

doodyparizada:

### DIFF
--- a/scripts/src/client.ts
+++ b/scripts/src/client.ts
@@ -145,7 +145,6 @@ export class Client {
 
 		const network = KinNetwork.from(
 			this.blockchainConfig.network_passphrase,
-			this.blockchainConfig.asset_issuer,
 			this.blockchainConfig.horizon_url);
 
 		const data = Client.normalizeSignInPayload(signInPayload);
@@ -320,14 +319,14 @@ export class Client {
 		}
 	}
 
-	public async submitOrder(orderId: string, content?: string, transaction?: string): Promise<Order> {
+	public async submitOrder(orderId: string, options: {content?: string, transaction?: string} = {}): Promise<Order> {
 		try {
 			const res = await this.requests
-				.request(`/v2/orders/${ orderId }`, { transaction })
+				.request(`/v2/orders/${ orderId }`, { transaction: options.transaction, content: options.content })
 				.post<Order>();
 			return res.data;
 		} catch (e) {
-			console.log(`error while submitting order ${ orderId } with content: "${ content }"`);
+			console.log(`error while submitting order ${ orderId } with options: "${ options }"`);
 			throw e;
 		}
 	}
@@ -375,7 +374,7 @@ export class Client {
 			throw new Error("first set a wallet");
 		}
 
-		await this.wallet.trustKin();
+		// NOP await this.wallet.trustKin();
 	}
 
 	public async logout() {
--- a/scripts/src/client.ts
+++ b/scripts/src/client.ts
@@ -61,21 +61,32 @@ export class ClientRequests {
 	public static async create(data: SignInPayload, headers?: StringMap) {
 		// get blockchain version for current app
 		const appId = this.extractAppId(data.jwt);
-		const blockchainVersion = (await axios.get<number>(
-				MARKETPLACE_BASE + `/v2/applications/${ appId }/blockchain_version`)
-		).data.toString() as BlockchainVersion;
-		console.log(`connected to app ${ appId } with version ${ blockchainVersion }`);
+		const blockchainVersion = await this.getBlockchainVersion(appId);
+		const token = await this.getToken(data.jwt, headers);
 
-		const res = await axios.post<{ auth: AuthToken; }>(MARKETPLACE_BASE + "/v2/users", {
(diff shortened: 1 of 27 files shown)


scripts/src/client.ts
```diff
@@ -320,18 +319,22 @@ export class Client {
 		}
 	}
 
-	public async submitOrder(orderId: string, content?: string): Promise<Order> {
+	public async submitOrder(orderId: string, options: {content?: string, transaction?: string} = {}): Promise<Order> {
```
nitzantomer: i'd do it like this:
```
public async submitOrder(orderId: string, options: { content: string } = {}): Promise<Order>;
public async submitOrder(orderId: string, options: { transaction: string } = {}): Promise<Order>;
public async submitOrder(orderId: string, options: { content?: string, transaction?: string } = {}): Promise<Order> {
    ...
}
```
that way when using this it's clear that i must pass either `{ content: "..." }` or `{ transaction: "..." }`

scripts/src/client.ts
```diff
@@ -320,18 +319,22 @@ export class Client {
...
 		try {
 			const res = await this.requests
-				.request(`/v2/orders/${ orderId }`, { content })
+				.request(`/v2/orders/${ orderId }`, { transaction: options.transaction, content: options.content })
```
nitzantomer: why not just:
```
.request(`/v2/orders/${ orderId }`, options)
```
?

scripts/src/client.ts
```diff
@@ -320,18 +319,22 @@ export class Client {
...
 				.post<Order>();
 			return res.data;
 		} catch (e) {
-			console.log(`error while submitting order ${ orderId } with content: "${ content }"`);
+			console.log(`error while submitting order ${ orderId } with options: "${ options }"`);
```
nitzantomer: that will stringify the `options`, this should be better:
```
console.log("error while submitting order ${ orderId } with options: ", options);
```

scripts/src/client.ts
```diff
@@ -371,7 +374,7 @@ export class Client {
 			throw new Error("first set a wallet");
 		}
 
-		await this.wallet.trustKin();
+		// NOP await this.wallet.trustKin();
```
nitzantomer: just remove it, no?

scripts/src/client.ts
```diff
@@ -62,27 +58,44 @@ export class ClientError extends Error {
...
 
-	public static async getConfig(): Promise<ConfigResponse> {
+	public static async getServerConfig(): Promise<ConfigResponse> {
 		const res = await axios.get<ConfigResponse>(MARKETPLACE_BASE + "/v2/config");
 		return res.data;
 	}
 
+	private static extractAppId(jwt: string) {
+		return (jsonwebtoken.decode(jwt) as { iss: string }).iss;
+
```
liorama: Extra line is extra

scripts/src/client.ts
```diff
@@ -62,27 +58,44 @@ export class ClientError extends Error {
...
 export class ClientRequests {
-	public static async create(data: any, headers?: StringMap) {
-		const res = await axios.post<{ auth: AuthToken; }>(MARKETPLACE_BASE + "/v2/users", data, { headers });
-		return new ClientRequests(res.data.auth);
+	public static async create(data: SignInPayload, headers?: StringMap) {
+		// get blockchain version for current app
+		const appId = this.extractAppId(data.jwt);
+		const blockchainVersion = (await axios.get<number>(
+				MARKETPLACE_BASE + `/v2/applications/${ appId }/blockchain_version`)
+		).data.toString() as BlockchainVersion;
```
liorama: Thid id real ugly, can you extract that the url into a const

scripts/src/client.ts
```diff
@@ -210,11 +208,12 @@ export class Client {
...
 
 	public async updateWallet(walletAddress?: string) {
+		const kinjs = this.requests.blockchainVersion === "2" ? kinjs1 : kinjs2;
 		const keys = !walletAddress ?
-			Keypair.random() :
+			kinjs.Keypair.random() :
 			(walletAddress.startsWith("S") ?
-				Keypair.fromSecret(walletAddress) :
-				Keypair.fromPublicKey(walletAddress));
+				kinjs.Keypair.fromSecret(walletAddress) :
```
liorama: How is `walletAddress` a secret?

scripts/src/client.ts
liorama: Extra line is extra

scripts/src/client.ts
liorama: These 3 lines are butt ugly, can you extact the url string into a const please

doodyparizada: I don't see why extracting the URL to a consts makes it nicer. The rest of the code uses inline urls when making requests. The second part of `.data.toString()` can be moved/ turned into a function maybe

scripts/src/client.ts
doodyparizada: when `walletAddress.startsWith("S")`

scripts/src/client.ts
doodyparizada: right!

PR #307, opened 2019-04-01
cache authtoken

doodyparizada:
provide a caching layer over the token to user resolution.
Once a token is resolved it will be stored in redis for 15 minutes before resolving again.

This should reduce DB calls for the same user in the same session.
*Note:* the need for a custom JSON.parse function for converting date strings to Date objects.

### DIFF
--- a/scripts/src/public/services/users.ts
+++ b/scripts/src/public/services/users.ts
@@ -2,7 +2,7 @@ import { Brackets } from "typeorm";
 
 import * as metrics from "../../metrics";
 import { Order } from "../../models/orders";
-import { dateParser, normalizeError, readUTCDate } from "../../utils/utils";
+import { dateParser, isNothing, normalizeError, readUTCDate } from "../../utils/utils";
 import { Application } from "../../models/applications";
 import { getDefaultLogger as logger } from "../../logging";
 import { User, AuthToken as DbAuthToken, WalletApplication } from "../../models/users";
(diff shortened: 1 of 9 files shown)


scripts/src/public/services/users.ts
```diff
@@ -139,8 +141,24 @@ export async function activateUser(authToken: DbAuthToken, user: User): Promise<
 }
 
 export async function userExists(appId: string, appUserId: string): Promise<boolean> {
+	// get from cache
+	const redis = getRedisClient();
+	const key = `app:${ appId }:user:${ appUserId }:exists`;
+	const value = await redis.async.get(key);
+	if (value) {
+		return JSON.parse(value);
```
nitzantomer: the function returns `Promise<boolean>`, but you return here the value.
shouldn't it just be `return true`?

nitzantomer (dismissed):

scripts/src/public/services/users.ts
doodyparizada: It's ok the function is `async` and this is syntactic sugar.
I'm also caching `false` responses but only for 15 minutes

nitzantomer: @doodyparizada 
what does the async part have to do with this?
my point is that `JSON.parse(value)` won't return a boolean, so it's not `Promise<boolean>` but `Promise<User>`, which is wrong according to the function signature.

PR #308, opened 2019-04-02
ECO-1253 don't return 500 on 404 from migration service and ECO-1263 …

liorama:
```v2/migration/info```:
* Don't throw 500 for a 404. Instead return ```should_migrate: false```
* Return is wallet allowed to be restored (wallet+appid)

- [x] linked to a JIRA issue

### DIFF
--- a/scripts/src/public/services/migration.ts
+++ b/scripts/src/public/services/migration.ts
@@ -5,6 +5,7 @@ import { getConfig } from "../config";
 
 import axios from "axios";
 import { getDefaultLogger as logger } from "../../logging";
+import { isRestoreAllowed } from "./users";
 
 const axiosRetry = require("axios-retry"); // TODO: this fails the tests: import axiosRetry from "axios-retry";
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: scripts/src/public/services/migration.ts [hunk 1]
```diff
@@ -15,20 +16,33 @@ axiosRetry(client, { retries: 6, retryCondition: () => true, shouldResetTimeout:
...
-
+	const publicAddress = req.params.public_address;
+	const appId = req.params.app_id;
+	logger().info(`handling account status request app_id: ${ appId } public_address: ${ publicAddress }`);
+
+	const app_blockchain_version = await getAppBlockchainVersion(appId);
+
+	const isBurnedRes = await client.get(`${ getConfig().migration_service }/accounts/${ publicAddress }/status`, {
+		validateStatus: status => {
+			return status < 500; // Reject only if the status code is greater than or equal to 500
```
doodyparizada: I think you can define this globally when defining the client

scripts/src/public/services/migration.ts
```diff
@@ -15,20 +16,33 @@ axiosRetry(client, { retries: 6, retryCondition: () => true, shouldResetTimeout:
...
+		}
+	});
+	if (isBurnedRes.status === 404) {
+		isBurnedRes.data.is_burned = true; // Wallet doesn't exist
+	} else if (isBurnedRes.status > 400) {
+		logger().info(`Something went wrong, migration service response`, isBurnedRes);
+		throw Error(`Something went wrong, migration service response: ${ JSON.stringify(isBurnedRes.data) }`);
```
doodyparizada: Throw a marketplace error. Any other error like the above will convert to 500

doodyparizada (changes requested):

doodyparizada (approved):

PR #309, opened 2019-04-04
Lior/eco 1233 support sdk version rules

liorama:
Different offer images for different versions

- [x] linked to a JIRA issue


scripts/src/public/services/offers.ts
```diff
@@ -44,12 +46,49 @@ type OfferTranslations = {
...
+	{
+		comparator: ">=1.0.0",
+		data: {
+			"https://cdn.kinecosystem.com/thumbnails/offers/quiz_4.png": "https://s3.amazonaws.com/assets.kinecosystemtest.com/images/offer_placeholder.png",
+			"https://cdn.kinecosystem.com/thumbnails/offers/earn_offer3.png": "https://s3.amazonaws.com/assets.kinecosystemtest.com/images/offer_placeholder.png",
+			"https://cdn.kinecosystem.com/thumbnails/offers/future_poll.png": "https://s3.amazonaws.com/assets.kinecosystemtest.com/images/offer_placeholder.png",
+			"https://cdn.kinecosystem.com/thumbnails/offers/marketplace_poll.png": "https://s3.amazonaws.com/assets.kinecosystemtest.com/images/offer_placeholder.png",
+			"https://cdn.kinecosystem.com/thumbnails/offers/kik_poll.png": "https://s3.amazonaws.com/assets.kinecosystemtest.com/images/offer_placeholder.png",
+			"https://cdn.kinecosystem.com/thumbnails/offers/quiz_1.png": "https://s3.amazonaws.com/assets.kinecosystemtest.com/images/offer_placeholder.png",
+		}
```
liorama: @doodyparizada @nitzantomer Change this according to the new images mapping.
I'm pretty sure we need to have clients move to `1.0.0` as the UI is a breaking change, but discuss/argue amongst yourselves and decide.

Note: Rules are evaluated and take precedence in order. I'll add a comment.

scripts/src/client.ts
```diff
@@ -129,19 +132,23 @@ export class ClientRequests {
...
 
+type ClientConfig = { headers?: StringMap, sdkVersion?: string };
+
 export class Client {
-	public static async create(signInPayload: SignInPayload, config?: { headers?: StringMap }): Promise<Client> {
+	public static async create(signInPayload: SignInPayload, config: ClientConfig = {}): Promise<Client> {
 		if (!this.blockchainConfig) {
 			this.blockchainConfig = (await ClientRequests.getConfig()).blockchain;
 		}
+		config.headers = config.headers || {};
```
doodyparizada: you don't need this line if the config defaults {}

docker-compose.yaml
```diff
@@ -32,6 +32,7 @@ services:
       APP_DB_HOST: postgres
       APP_DB_DATABASE: ecosystem
       NODEMON_WATCH_DIRS: ${NODEMON_WATCH_DIRS}
+      CACHE_TTL: '{ default: 0 }'
```
doodyparizada: This isn't valid json (missing quotes around default)

scripts/src/client.ts
liorama: But `headers` might be missing. It's a ramanant that's left after some refactoring, I can change the default to `{headers:{}}` now.

scripts/src/public/services/offers.ts
doodyparizada: We left this to you. There was some confusion of the current location of the images

doodyparizada:
Failing on `(because: Cannot read property 'join' of undefined)` Which I remember I encountered and fixed on kin.js. I don't remember the cause though. maybe the testing wallet was accidentally burned, so we'll need to replace it. You can take the current kin2 wallet from our kin3 branch

PR #313, opened 2019-04-07
Fix cross-app restore bug

liorama:
- [ ] linked to a JIRA issue


doodyparizada (approved):
It's the best if we have a test for this

PR #320, opened 2019-04-14
only create wallet when not created

doodyparizada:
Requires:
```
ALTER TABLE user_wallets ADD COLUMN account_created_date TIMESTAMP NULL;
```

Depends on https://github.com/kinecosystem/payment-service/pull/42

## Description
Only create wallet on blockchain if it hasn't been created yet. When the wallet is created, we mark this in the DB so it won't be created next time.

### DIFF
--- a/scripts/src/internal/services.ts
+++ b/scripts/src/internal/services.ts
@@ -1,7 +1,7 @@
 import { getDefaultLogger as logger } from "../logging";
 import * as metrics from "../metrics";
 import * as db from "../models/orders";
-import { User } from "../models/users";
+import { User, Wallet } from "../models/users";
 import { pick, removeDuplicates } from "../utils/utils";
 import { Asset, Offer, OrderValue } from "../models/offers";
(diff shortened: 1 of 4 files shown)


scripts/src/internal/services.ts
```diff
@@ -18,24 +18,33 @@ import { Application, AppOffer } from "../models/applications";
 
 const BLOCKCHAIN = "kin-prod";
 const RS512_APPS = ["test", "smpl"];
+const ACCOUNT_EXISTS_REASON = "account exists";
```
liorama: If this is the blockchain error message I would rename this to: `ACCOUNT_EXISTS_BLOCKCHAIN_ERROR` because it's not obvious what this is and where it originates from.

doodyparizada: It's the message that the payment-service sends when the reason why account creation failed is that the account already exists. We currently send a single string as the reason. A better solution would be to send a code (strings can change). But adding this extra piece of info feels out of scope

PR #322, opened 2019-04-28
Eco 1367 5xx error fix

doodyparizada:


.travis.yml
```diff
@@ -6,18 +6,18 @@ node_js:
...
   - make create-jwt-keys
   - make build-image
-  - make db-docker
+  - make build db
   - NODEMON_WATCH_DIRS="-w /no_such_dir" make up
 
 script:
   - make test
   - ./node_modules/codecov/bin/codecov
+  - make clear-db db
```
doodyparizada: required to reset the DB after running the unittests

Makefile
```diff
@@ -69,6 +65,7 @@ psql:
 redis-cli:
 	docker-compose -f docker-compose.yaml -f docker-compose.deps.yaml -f docker-compose.tests.yaml run --rm redis-cli
 
+db: db-docker
```
doodyparizada: use postgres for unittests locally too

docker-compose.yaml
```diff
@@ -74,7 +74,7 @@ services:
   postgres:
     image: postgres:10.4
     ports:
-      - 5432
+      - 25432:5432
```
doodyparizada: hardcode the port mapping so it can be used in the test/config.json

scripts/src/models/orders.ts
```diff
@@ -155,7 +155,7 @@ export const Order = {
 		const results: Array<{ offerId: string, cnt: number }> = await query.getRawMany();
 		const map = new Map<string, number>();
 		for (const res of results) {
-			map.set(res.offerId, res.cnt);
+			map.set(res.offerId, Number(res.cnt));
```
doodyparizada: unit tests uncovered a bug when running in postgres mode

scripts/src/models/orders.ts
```diff
@@ -240,24 +240,22 @@ export const Order = {
 	 * @param      {number}  limit
 	 * @return     {Promise<T[]>}  filtered orders including p2p
 	 */
-	async getAll<T extends Order>(filters: GetOrderFilters & { userId?: string }, limit?: number): Promise<T[]> {
-		const allOrders = await this.genericGet(filters).getMany(); // can be replaced by cache
-
-		const ids: string[] = allOrders.map(order => order.id);
+	async getAll<T extends Order>(filters: GetOrderFilters & ({ userId: string } | { walletAddress: string }), limit?: number): Promise<T[]> {
+		const contexts = await OrderContext.find(filters.userId ? { userId: filters.userId } : { wallet: filters.walletAddress }); // can be replaced by cache
```
doodyparizada: This query is easier to cache and "should" be easier on the DB .
The filters will be applied in the following query

scripts/src/models/orders.ts
```diff
@@ -240,24 +240,22 @@ export const Order = {
...
 		// remove all context related fields to get the second context
 		delete filters.userId;
 		delete filters.walletAddress;
+
 		const userOrdersQuery = this.genericGet(filters)
 			.andWhere(`ordr.id IN (:ids)`, { ids });
 
 		if (limit) {
-			userOrdersQuery.limit(limit);
+			userOrdersQuery.take(limit);
```
doodyparizada: using `.take(...)` instead of `.limit(...)` cases typeORM to be careful with relations when implementing pagination. so no more p2p order with only one context (it implements this though by issuing 2 queries instead of 1)

scripts/src/public/routes/orders.ts
```diff
@@ -138,7 +138,7 @@ export type GetOrderHistoryRequest = AuthenticatedRequest & {
 export const getOrderHistory = async function(req: GetOrderHistoryRequest, res: Response) {
 	const filters = {
 		origin: req.query.origin,
-		offerId: req.query.offer_id
+		offerId: req.query.offer_id || req.query.offerId // old clients send offerId instead of offer_id
```
doodyparizada: This was fixed by in the Android code a while ago, but not all partners have adopted the new version

scripts/src/public/services/users.ts
```diff
@@ -101,8 +101,15 @@ export async function isRestoreAllowed(walletAddress: string, appId: string, add
 		}
 	} else if (addOnNonExisting) {
 		logger().info(`Wallet ${ walletAddress } does not exist in wallet_application table, add`);
-		const newWallet = WalletApplication.create({ walletAddress, appId });
-		await newWallet.save();
+		try {
+			const newWallet = WalletApplication.create({ walletAddress, appId });
+			await newWallet.save();
+		} catch (e) {
+			// maybe caught a "violates unique constraint" error, check by finding the wallet again
```
liorama: // might catch a...

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "ES6",
+    "target": "ESnext",
```
liorama: This might introduce feature not support by node. It's fine as long as we always test new tsc versions with our production node version (which we do in CI/CD) but we do need to be aware of this. If it's just for async/await then ES2017 should surfice.

liorama:
@doodyparizada remove SQLite from package.json

liorama (approved):

PR #323, opened 2019-04-29
ECO-1237 retry migration service requests only for network/500 errors

liorama:


doodyparizada (approved):
too long please split into 10 PRs

PR #337, opened 2019-05-25
rate limit user requests

doodyparizada:


scripts/src/config.ts
```diff
@@ -7,7 +7,9 @@ import { LogTarget } from "./logging";
 
 export interface LimitConfig {
 	hourly_registration: number;
-	minute_registration: number;
+	a: number;
```
liorama: ?

scripts/src/utils/rate_limit.ts
```diff
@@ -109,3 +109,18 @@ export async function assertRateLimitRegistration(app: Application) {
...
+	return await checkRateLimit(`user_reqs:${ userId }:${ duration.asSeconds() }`, duration, limit, TooManyUserRequests,
+		1, 2); // use lower number of buckets to reduce stress on redis as this is called on each user request
+}
+
+export async function assertRateLimitUserRequests(user: User) {
+	const app = (await Application.get(user.appId))!;
+
+	const limiters = [
+		await checkRateLimitUserRequests(app.id, app.config.limits.hourly_user_requests || 150, moment.duration({ hours: 1 })),
+		await checkRateLimitUserRequests(app.id, app.config.limits.minute_user_requests || 20, moment.duration({ minutes: 1 }))
```
liorama: I think you meant to pass the user ID for both calls to `checkRateLimitUserRequests`

liorama (changes requested):

scripts/src/config.ts
doodyparizada: hahaha

scripts/src/utils/rate_limit.ts
doodyparizada: you are right!

Doc999tor (approved):

PR #338, opened 2019-05-25
use precise query to get a single order

doodyparizada:
`getOne` will also find by offerId+nonce+userId
`getAll` will filter out the original orderId list with the given filter (instead of ignoring the filter).

### DIFF
--- a/scripts/src/models/orders.ts
+++ b/scripts/src/models/orders.ts
@@ -241,9 +241,8 @@ export const Order = {
 	 */
 	async getAll<T extends Order>(filters: GetOrderFilters & ({ userId: string } | { walletAddress: string }), limit: number = 25): Promise<T[]> {
 		const orderIdQuery = await (this.genericGet(filters)
-			.limit(limit + 1)
-			.getMany()); // +1 to account for p2p
-		console.log("GOT", orderIdQuery);
+			.limit(limit)
+			.getMany());
 		const ids = orderIdQuery.map(res => res.id);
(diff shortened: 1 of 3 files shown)


scripts/src/models/orders.ts
```diff
@@ -240,9 +239,12 @@ export const Order = {
 	 * @param      {number}  limit
 	 * @return     {Promise<T[]>}  filtered orders including p2p
 	 */
-	async getAll<T extends Order>(filters: GetOrderFilters & ({ userId: string } | { walletAddress: string }), limit?: number): Promise<T[]> {
-		const contexts = await OrderContext.find(filters.userId ? { userId: filters.userId } : { wallet: filters.walletAddress }); // can be replaced by cache
-		const ids = contexts.map(c => c.orderId);
+	async getAll<T extends Order>(filters: GetOrderFilters & ({ userId: string } | { walletAddress: string }), limit: number = 25): Promise<T[]> {
+		const orderIdQuery = await (this.genericGet(filters)
+			.limit(limit + 1)
+			.getMany()); // +1 to account for p2p
```
liorama: Move the comment to the right line, and maybe elaborate more, not sure I understand the reasoning.

liorama (dismissed):

scripts/src/models/orders.ts
doodyparizada: It was irrelevant that comment and the +1. I removed it and tested it in a unittest

PR #339, opened 2019-05-30
Support zero balance non-burn migration

doodyparizada:
Already checked this with the new migration-service.
zero balance and non zero balance
from app start and while using app.
~need to QA what happens when balance is 0 but kin3 account doesn't exist~ - done

### DIFF
--- a/scripts/src/internal/services.ts
+++ b/scripts/src/internal/services.ts
@@ -30,7 +30,6 @@ export type WalletCreationSuccessData = {
 export async function walletCreationSuccess(data: WalletCreationSuccessData) {
 	createStellarAccountCreationSucceeded(data.id).report();
 	logger().info("wallet created", data);
-	await updateWalletCreationDate(data.wallet_address, "createdDateKin2");
 }
 
 export type WalletCreationFailureData = {
@@ -227,15 +226,5 @@ export async function initPaymentCallbacks(): Promise<Watcher> {
 }
 
 export async function markWalletBurnt(walletAddress: string) {
-	await updateWalletCreationDate(walletAddress, "createdDateKin3");
-}
-
-async function updateWalletCreationDate(walletAddress: string, createdDateField: "createdDateKin2" | "createdDateKin3") {
-	const wallet = await WalletApplication.findOne({ walletAddress });
-	if (!wallet) {
-		throw NoSuchWallet(walletAddress);
-	}
-
-	wallet[createdDateField] = new Date();
-	await wallet.save();
+	await WalletApplication.updateCreatedDate(walletAddress, "createdDateKin3");
 }
--- a/scripts/src/public/app.ts
+++ b/scripts/src/public/app.ts
@@ -5,7 +5,6 @@ import { getConfig } from "./config";
 import { init as initModels } from "../models/index";
 import { init as initRemoteConfig } from "./routes/config";
 import { init as initMigration } from "./services/migration";
-
 import { createRoutes, createV1Routes } from "./routes/index";
 import { generalErrorHandler, init as initCustomMiddleware, notFoundHandler } from "./middleware";
 
--- a/scripts/src/public/app.ts
+++ b/scripts/src/public/app.ts
@@ -4,6 +4,8 @@ import { initLogger } from "../logging";
 import { getConfig } from "./config";
 import { init as initModels } from "../models/index";
 import { init as initRemoteConfig } from "./routes/config";
+import { init as initMigration } from "./services/migration";
+
 import { createRoutes, createV1Routes } from "./routes/index";
 import { generalErrorHandler, init as initCustomMiddleware, notFoundHandler } from "./middleware";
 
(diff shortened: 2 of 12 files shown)


scripts/src/internal/services.ts
```diff
@@ -227,15 +226,5 @@ export async function initPaymentCallbacks(): Promise<Watcher> {
...
-
-async function updateWalletCreationDate(walletAddress: string, createdDateField: "createdDateKin2" | "createdDateKin3") {
-	const wallet = await WalletApplication.findOne({ walletAddress });
-	if (!wallet) {
-		throw NoSuchWallet(walletAddress);
-	}
-
-	wallet[createdDateField] = new Date();
-	await wallet.save();
+	await WalletApplication.updateCreatedDate(walletAddress, "createdDateKin3");
```
doodyparizada: Change to single update instead of select+update

scripts/src/internal/services.ts
```diff
@@ -30,7 +30,6 @@ export type WalletCreationSuccessData = {
 export async function walletCreationSuccess(data: WalletCreationSuccessData) {
 	createStellarAccountCreationSucceeded(data.id).report();
 	logger().info("wallet created", data);
-	await updateWalletCreationDate(data.wallet_address, "createdDateKin2");
```
doodyparizada: This location was wrong for this update. The callback is called from both kin2 and kin3 payment service and it always updates kin2 creation date.
I change it to update the creation date to just before requesting the createWallet request on either kin2 or kin3.

OrenZak: ok

scripts/src/public/app.ts
```diff
@@ -4,6 +4,8 @@ import { initLogger } from "../logging";
 import { getConfig } from "./config";
 import { init as initModels } from "../models/index";
 import { init as initRemoteConfig } from "./routes/config";
+import { init as initMigration } from "./services/migration";
+
```
OrenZak: remove

scripts/src/public/auth.ts
```diff
@@ -4,6 +4,7 @@ import * as httpContext from "express-http-context";
 import { dateParser, Mutable } from "../utils/utils";
 import { AuthToken, User } from "../models/users";
 import { getRedisClient } from "../redis";
+
```
OrenZak: ?

scripts/src/public/services/migration.ts
```diff
@@ -2,18 +2,77 @@ import * as express from "express";
...
 	app_blockchain_version: BlockchainVersion;
 	restore_allowed: boolean;
 };
 
+export async function init() {
+	BLOCKCHAIN = await getBlockchainConfig("2");
+	BLOCKCHAIN3 = await getBlockchainConfig("3");
+}
+
+async function getKin2Balance(walletAddress: string) {
```
OrenZak: rename to `hasKin2Balance` ? and add some docs

scripts/src/public/services/migration.ts
```diff
@@ -2,18 +2,77 @@ import * as express from "express";
...
+	try {
+		await httpClient.get<WalletResponse>(`${ BLOCKCHAIN3.horizon_url }/accounts/${ walletAddress }`);
+		return true;
+	} catch (e) {
+		return false;
+	}
+}
+
+// returns true if migration call succeeded
+async function migrateZeroBalance(walletAddress: string) {
```
OrenZak: throw an exception if didn't success ?

scripts/src/public/services/migration.ts
```diff
@@ -22,15 +81,30 @@ export const accountStatus = async function(req: express.Request, res: express.R
...
+	if (shouldMigrate
+		&& wallet
+		// zero balance accounts don't need to migrate.
+		// We check the balance of this account on kin2 prior to deciding
+		&& await getKin2Balance(wallet.walletAddress) === 0
+		// account must be created on kin3, otherwise the migrate call will block for a few seconds,
+		// which we prefer the client will do
+		&& await hasKin3Account(wallet.walletAddress)) {
+		try {
+			await migrateZeroBalance(wallet.walletAddress);
```
OrenZak: make sure you throw an exception

scripts/src/public/services/migration.ts
```diff
@@ -22,15 +81,30 @@ export const accountStatus = async function(req: express.Request, res: express.R
...
+			shouldMigrate = false;
+		} catch (e) {
+			// fail to call migrate - let user call it instead
+			shouldMigrate = true;
+		}
+	}
```
OrenZak: Check also if it's ok with iOS implementation.

scripts/src/public/services/users.ts
```diff
@@ -114,6 +114,23 @@ export async function isRestoreAllowed(walletAddress: string, appId: string, add
 	return true;
 }
 
+async function createWallet(walletAddress: string, user: User) {
+	logger().info(`creating stellar wallet for user ${ user.appId }: ${ walletAddress }`);
```
OrenZak: maybe add also which blockchain ver

OrenZak (dismissed):
👍

PR #342, opened 2019-06-05
Lior/rename order query aliases

liorama:
So it'll be easier for us to recognize some of our queries when looking at Aurora's performance insights.

- [ ] linked to a JIRA issue

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -214,9 +214,9 @@
 			"dev": true
 		},
 		"@types/node": {
-			"version": "10.12.12",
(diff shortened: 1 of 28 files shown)


doodyparizada:
I don't see how this helps you distinguish the queries any better than just the queries themselves. This just adds weird names as the alias. All the queries you changed have look different.
Would be more interesting to have a different source per genericGet query builder

liorama:
Passed Unit tests and system tests (locally)

PR #345, opened 2019-06-12
add bulk creation of users endpoint

liorama:
- [x] linked to a JIRA issue

### DIFF
--- a/scripts/src/errors.ts
+++ b/scripts/src/errors.ts
@@ -26,7 +26,8 @@ const CODES = {
 			ExpiredJwt: 9,
 			InvalidJwtIssuedTime: 10,
 			MissingFieldJWT: 11,
-			BadJWTInput: 12
+			BadJWTInput: 12,
+			InvalidJwtField: 13,
 		}
 	},
 	Unauthorized: {
@@ -37,6 +38,7 @@ const CODES = {
 			InvalidApiKey: 3,
 			TOSMissingOrOldToken: 4,
 			CrossAppWallet: 5,
+			BulkUserCreation: 6,
 		}
 	},
 	NotFound: {
@@ -153,7 +155,11 @@ export function TOSMissingOrOldToken() {
 }
 
 export function CrossAppWallet(wallet: string, app: string) {
-	return UnauthorizedError("CrossAppWallet", `Wallet ${wallet} does not belong to current app ${app}`);
+	return UnauthorizedError("CrossAppWallet", `Wallet ${ wallet } does not belong to current app ${ app }`);
+}
+
+export function BulkUserCreation(app: string, requestedAmount: number, allowedNumber: number) {
+	return UnauthorizedError("BulkUserCreation", `Bulk creation of ${ allowedNumber } is allowed for app ${ app }, ${ requestedAmount } requested`);
 }
 
 function NotFoundError(key: keyof typeof CODES.NotFound.types, message: string) {
@@ -197,7 +203,7 @@ function ConflictError(key: keyof typeof CODES.Conflict.types, message: string) 
 }
 
 export function ExternalOrderAlreadyCompleted(orderId: string, status: string) {
-	const error = ConflictError("ExternalOrderAlreadyCompleted", `Can't create order for offer. Another order is ${status}`);
+	const error = ConflictError("ExternalOrderAlreadyCompleted", `Can't create order for offer. Another order is ${ status }`);
 	error.setHeader("Location", `/v1/orders/${ orderId }`);
 	return error;
 }
@@ -267,7 +273,11 @@ export function InvalidJwtIssuedTime(iat: number) {
 }
 
 export function MissingFieldJWT(fieldName: string) {
-	return BadRequestError("MissingFieldJWT", `The JWT ${ fieldName } field is missing`);
+	return BadRequestError("MissingFieldJWT", `The JWT ${ fieldName } field is missing or invalid`);
+}
+
+export function InvalidJwtField(message: string) {
+	return BadRequestError("InvalidJwtField", message);
 }
 
 export function BadJWTInput(token: string) {
--- a/scripts/src/errors.ts
+++ b/scripts/src/errors.ts
@@ -273,7 +273,7 @@ export function InvalidJwtIssuedTime(iat: number) {
 }
 
 export function MissingFieldJWT(fieldName: string) {
-	return BadRequestError("MissingFieldJWT", `The JWT ${ fieldName } field is missing or invalid`);
+	return BadRequestError("MissingFieldJWT", `The JWT ${ fieldName } field is missing`);
 }
 
 export function InvalidJwtField(message: string) {
--- a/scripts/src/public/app.ts
+++ b/scripts/src/public/app.ts
@@ -5,7 +5,7 @@ import { getConfig } from "./config";
 import { init as initModels } from "../models/index";
 import { init as initRemoteConfig } from "./routes/config";
 import { init as initMigration } from "./services/migration";
-import { createRoutes, createV1Routes } from "./routes/index";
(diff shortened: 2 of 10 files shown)


scripts/src/errors.ts
```diff
@@ -26,7 +26,8 @@ const CODES = {
 			ExpiredJwt: 9,
 			InvalidJwtIssuedTime: 10,
 			MissingFieldJWT: 11,
-			BadJWTInput: 12
+			BadJWTInput: 12,
+			InvalidJwtField: 13,
```
doodyparizada: Haha I just added this error myself with the same name and same number

scripts/src/errors.ts
```diff
@@ -267,7 +273,11 @@ export function InvalidJwtIssuedTime(iat: number) {
 }
 
 export function MissingFieldJWT(fieldName: string) {
-	return BadRequestError("MissingFieldJWT", `The JWT ${ fieldName } field is missing`);
+	return BadRequestError("MissingFieldJWT", `The JWT ${ fieldName } field is missing or invalid`);
```
doodyparizada: Why also invalid?

scripts/src/public/app.ts
```diff
@@ -32,6 +32,7 @@ export const app: express.Express = createApp();
 // routes
 createRoutes(app, "/v2");
 createV1Routes(app, "/v1");
+createPublicFacing(app, "/partners/v1");
```
doodyparizada: I don't like that the version is not the first thing in the URL. You can just use `/v1/applications/<>/...` or `v2`

scripts/src/public/routes/index.ts
```diff
@@ -102,5 +102,15 @@ export function createV1Routes(app: express.Express, pathPrefix?: string) {
...
+
+export function createPublicFacing(app: express.Express, pathPrefix?: string) {
+	function prefix(path: string): string {
+		if (!pathPrefix) {
+			return path;
+		}
+		return `${ pathPrefix }/${ path }`;
+	}
+
+	app.post(prefix("users/bulk"), bulkUserCreation);
```
doodyparizada: Maybe user-bulk?

scripts/src/public/routes/users.ts
```diff
@@ -219,3 +224,49 @@ export const logoutUser = async function(req: AuthenticatedRequest, res: Respons
...
+export const bulkUserCreation = async function(req: BulkUserCreationRequest, res: Response) {
+	const jwtList = req.body.user_data as Array<[string, string]>;
+	const firstJwt = jsonwebtoken.decode(jwtList[0][0], { complete: true }) as JWTContent<Partial<RegisterPayload>, "register">; // Use the app id from the first JWT
+	const app_id = firstJwt.payload.iss;
+	logger().info(`bulkUserCreation for ${ app_id }`);
+	const app = await Application.get(app_id);
+	if (!app) {
+		throw NoSuchApp(app_id);
+	}
+	let allowedCreations: number = app.config.bulk_user_creation_allowed || 0;
```
doodyparizada: `:number` is redundent

scripts/src/public/routes/users.ts
```diff
@@ -219,3 +224,49 @@ export const logoutUser = async function(req: AuthenticatedRequest, res: Respons
...
+	if (!allowedCreations || allowedCreations < jwtList.length) {
```
doodyparizada: Why not create up to the limit?

scripts/src/public/routes/users.ts
```diff
@@ -219,3 +224,49 @@ export const logoutUser = async function(req: AuthenticatedRequest, res: Respons
...
+		throw BulkUserCreation(app.id, jwtList.length, allowedCreations);
+	}
+	await app.save();
```
doodyparizada: why?

scripts/src/public/routes/users.ts
```diff
@@ -219,3 +224,49 @@ export const logoutUser = async function(req: AuthenticatedRequest, res: Respons
...
+	res.write(`requestId: ${ httpContext.get("reqId") }\n`);
+	await batch(jwtList, app.config.limits.minute_registration, 61 * 1000, async (sublist: Array<[string, string]>, firstIndexOfChunk) => {
+			await Promise.all(sublist.map(async ([jwt, publicAddress], index) => {
```
doodyparizada: too many spaces

scripts/src/public/routes/users.ts
```diff
@@ -219,3 +224,49 @@ export const logoutUser = async function(req: AuthenticatedRequest, res: Respons
...
+	if (!app) {
+		throw NoSuchApp(app_id);
+	}
+	let allowedCreations: number = app.config.bulk_user_creation_allowed || 0;
+	if (!allowedCreations || allowedCreations < jwtList.length) {
+		throw BulkUserCreation(app.id, jwtList.length, allowedCreations);
+	}
+	await app.save();
+	res.write(`requestId: ${ httpContext.get("reqId") }\n`);
+	await batch(jwtList, app.config.limits.minute_registration, 61 * 1000, async (sublist: Array<[string, string]>, firstIndexOfChunk) => {
```
doodyparizada: what is 61*1000

scripts/src/utils/utils.ts
```diff
@@ -187,3 +187,21 @@ export function dateParser(key: string, value: string) {
 
 	return value;
 }
+
+export async function batch(list: any[], chunkSize: number, delay: number, chunkCb: (chunk: any[], firstIndexOfChunk: number) => Promise<void>) {
```
doodyparizada: This whole function isn't inline with the rest of the coding standards in marketplace - i.e. using async/await instead of then for promises.
Also you have `console.log` and `setTimeout` which is very weird

scripts/src/public/routes/users.ts
```diff
@@ -219,3 +224,49 @@ export const logoutUser = async function(req: AuthenticatedRequest, res: Respons
...
+			await Promise.all(sublist.map(async ([jwt, publicAddress], index) => {
+				const currentItem = firstIndexOfChunk + (index + 1);
+				const context = await validateRegisterJWT(jwt);
+				const deviceId = context.deviceId;
+				if (context.appId !== app_id) {
+					throw InvalidJwtField("issuer (iss) fields of all supplied JWT must match");
```
doodyparizada: I would validate the whole bunch of JWTs and then only apply the user creation if all are valid

scripts/src/public/routes/users.ts
```diff
@@ -219,3 +224,49 @@ export const logoutUser = async function(req: AuthenticatedRequest, res: Respons
...
+	}
+	let allowedCreations: number = app.config.bulk_user_creation_allowed || 0;
+	if (!allowedCreations || allowedCreations < jwtList.length) {
+		throw BulkUserCreation(app.id, jwtList.length, allowedCreations);
+	}
+	await app.save();
+	res.write(`requestId: ${ httpContext.get("reqId") }\n`);
+	await batch(jwtList, app.config.limits.minute_registration, 61 * 1000, async (sublist: Array<[string, string]>, firstIndexOfChunk) => {
+			await Promise.all(sublist.map(async ([jwt, publicAddress], index) => {
+				const currentItem = firstIndexOfChunk + (index + 1);
```
doodyparizada: This calculation seems very confusing.
I think using a simple for loop and calling a function on a slice of users would look much more readable than using a callback chain.

Something like:
```
for (i = 0; i < jwt.length; i += BATCH) {
  slice = jwt.slice(i, i+BATCH);
  registerBatch(slice);
}
```

scripts/src/utils/utils.ts
doodyparizada: see https://github.com/kinecosystem/marketplace-server/pull/345/files#r292905889

PR #346, opened 2019-06-16
get last used wallet if no deviceId is given

doodyparizada:
@Elaz @liorama 

The iOS client relies on us giving the correct wallet address. When the deviceId changes on the client this caused the server to return `null` as the wallet. This will return the lastUsed wallet across all devices.

### DIFF
--- a/scripts/src/models/users.ts
+++ b/scripts/src/models/users.ts
@@ -45,14 +45,19 @@ export class User extends CreationDateModel {
 
 		if (deviceId) {
 			conditions.deviceId = deviceId;
+			const wallets = await Wallet.find(conditions);
+			if (wallets.length) {
+				return new Wallets(wallets);
+			} // else delete deviceId and let find generic wallets
+			delete conditions.deviceId;
 		}
 
-		const wallets = await Wallet.find(conditions);
+		let wallets = await Wallet.find(conditions);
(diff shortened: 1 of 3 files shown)


liorama (dismissed):

scripts/src/models/users.ts
```diff
@@ -45,14 +45,19 @@ export class User extends CreationDateModel {
...
+				return new Wallets(wallets);
+			} // else delete deviceId and let find generic wallets
+			delete conditions.deviceId;
 		}
 
-		const wallets = await Wallet.find(conditions);
+		let wallets = await Wallet.find(conditions);
 		if (wallets.length === 0 && this.walletAddress) {
-			await this.lazyMigrateWallet(deviceId); // TODO can we get rid of this?
+			await this.lazyMigrateWallet(deviceId);
```
liorama: There might not be a `deviceId` and then the next line will also return 0 wallets. Isn't it better to just return `this.walletAddress`?

doodyparizada: `lazyMigrateWallet` takes the deviceId from the token and moves users.wallet_ddress to user_wallets.
This function returns a Wallets object, not a string

liorama (approved):

PR #358, opened 2019-07-04
Victor/eco 1508/ordering column added to app offers

victordoc:
- [ ] linked to a JIRA issue

`alter table application_offers add column ordering int;` - will add the column to the prod db, I'll populate the new column manually
`create unique index app_ordering_uniq_idx on application_offers (app_id, ordering)` - unique index for preventing duplicates and sorting

### DIFF
--- a/scripts/src/models/applications.ts
+++ b/scripts/src/models/applications.ts
@@ -99,8 +99,7 @@ export class AppOffer extends BaseEntity {
 				.leftJoinAndSelect("app_offer.app", "app")
 				.where("app_offer.appId = :appId", { appId })
 				.andWhere("offer.type = :type", { type })
-				.orderBy("offer.amount", type === "earn" ? "DESC" : "ASC")
-				.addOrderBy("offer.id", "ASC")
+				.orderBy("offer.ordering", "ASC")
 				.getMany();
 			localCache.set(cacheKey, appOffers);
 		}
--- a/scripts/src/models/applications.ts
+++ b/scripts/src/models/applications.ts
@@ -108,6 +108,14 @@ export class AppOffer extends BaseEntity {
 		return appOffers;
 	}
 
+	public static async generate(appId: string, offerId: string, cap: Cap, walletAddress: string): Promise<AppOffer> {
+		const lastAppOffer = await AppOffer.findOne({ where: { appId }, order: { ordering: "DESC" } });
+		const orderingBufferStep = 10;
+		const lastAppOfferOrdering = (lastAppOffer && lastAppOffer.ordering) ? Number(lastAppOffer.ordering) : 0;
+		const newAppOfferOrdering = lastAppOfferOrdering + 1 * orderingBufferStep;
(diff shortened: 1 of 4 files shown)


doodyparizada:
In getOffers you should use the new ordering column for the order

scripts/src/models/applications.ts
```diff
@@ -99,15 +99,22 @@ export class AppOffer extends BaseEntity {
 				.leftJoinAndSelect("app_offer.app", "app")
 				.where("app_offer.appId = :appId", { appId })
 				.andWhere("offer.type = :type", { type })
-				.orderBy("offer.amount", type === "earn" ? "DESC" : "ASC")
-				.addOrderBy("offer.id", "ASC")
+				.orderBy("offer.ordering", "ASC")
```
doodyparizada: maybe still leave the other orderBy statements as secondary, so when you deploy to production you can leave `ordering` empty to begin with. In any case add the alter table statement here in the PR so we can review it too

doodyparizada (changes requested):

scripts/src/models/applications.ts
victordoc: What I thought is to alter the table first: add a column, populate it with data and afterwards to deploy this fix
How to add the `alter table statement` code to the PR? By a comment in the code or comment in the PR itself?

doodyparizada: just comment in the PR

victordoc: Done
I added a unique index to the column for performance and consistency

doodyparizada (approved):

scripts/src/models/applications.ts
```diff
@@ -99,15 +99,22 @@ export class AppOffer extends BaseEntity {
...
 		}
 
 		return appOffers;
 	}
 
+	public static async generate(appId: string, offerId: string, cap: Cap, walletAddress: string): Promise<AppOffer> {
+		const lastAppOffer = await AppOffer.findOne({ where: { appId }, order: { ordering: "DESC" } });
+		const orderingBufferStep = 10;
+		const lastAppOfferOrdering = (lastAppOffer && lastAppOffer.ordering) ? Number(lastAppOffer.ordering) : 0;
+		const newAppOfferOrdering = lastAppOfferOrdering + 1 * orderingBufferStep;
```
liorama: I think "OrderIndex" is a better suffix than Ordering, eg:
`newAppOfferOrderIndex`
(This applies to all relevant var names in this context)

scripts/src/models/applications.ts
```diff
@@ -120,6 +127,9 @@ export class AppOffer extends BaseEntity {
 	@Column({ name: "wallet_address" })
 	public walletAddress!: string;
 
+	@Column({ name: "ordering", type: "int" })
+	public ordering!: number;
```
liorama: Here also, I'd do `orderIndex`

PR #363, opened 2019-07-09
K8S and Jenkins support

ittiel:
- [ ] linked to a JIRA issue


liorama:
@ittiel @vanessatk so what are we doing with this?

PR #365, opened 2019-07-16
Remove wallet app cache

doodyparizada:


scripts/src/public/services/orders.ts
```diff
@@ -353,7 +353,9 @@ export async function submitOrder(
 	if (order.isMarketplaceOrder()) {
 		await submitFormAndMutateMarketplaceOrder(order, form);
 	}
-	const blockchainVersion = await WalletApplication.getBlockchainVersion(walletAddress);
+
+	const app = (await Application.all()).get(context.user.appId)!;
+	const blockchainVersion = app.config.blockchain_version === "3" ? "3" : await WalletApplication.getBlockchainVersion(walletAddress);
```
liorama: Add this in all places where you addWatcher

liorama: Also, make sure to pass `blockchainVersion` to `payment.payTo` instead of doing this agian there.

PR #368, opened 2019-07-23
use non mocking redis

doodyparizada:


scripts/src/tests/utils.spec.ts
```diff
@@ -275,4 +275,22 @@ describe("util functions", () => {
...
+				throw new Error("shouldn't reach 2");
+			}
+			await delay(500);
+			num -= 1;
+		}
+
+		const resource = `uniq:${ generateId() }`;
+		await Promise.all([
+			lock(resource, f),
+			lock(resource, f)]);
```
liorama: But if the delay is 500ms and the `RedLock` config is set to 10 retries with 200ms wait between them that means that the resource will be unlocked before all retries have been exhausted and the second call to `f` will be done and `num` increased to 2.

PR #370, opened 2019-07-31
Shaharsol/eco 1574/tranlations csv by title

shaharsol:
partial implementation of the suggested in the ticket. Didn't go as far as changing the db schema, just made sure template file read/write with be with offer.name as part of the key instead of offer.id. db offer_content and offer_content_translation still use offer_id


scripts/src/admin/translations.ts
```diff
@@ -123,18 +131,19 @@ async function getCsvRowData() {
 	let rows: CsvRow[] = [];
 	allOffers.forEach(offer => {
 		const offerId = offer.id;
+		const offerName = offer.name;
 		const offerContent: OfferContent = allContent.filter(obj => obj.offerId === offerId)[0];
```
liorama: You can change this to check for name and remove the `offerId` var completely.

scripts/src/admin/translations.ts
```diff
@@ -9,6 +9,8 @@ import { ContentType, Offer, OfferContent } from "../models/offers";
 import { OfferTranslation } from "../models/translations";
 import { path } from "../utils/path";
 
+import * as _ from "lodash";
```
liorama: I really don't see why you need this, please reconsider.

scripts/src/admin/translations.ts
```diff
@@ -266,37 +275,45 @@ async function insertIntoDb(data: OffersTranslation, language: string) {
...
-		} else {
-			const evalString = `offerTranslations.content.${ jsonPath }=translation`;
-			try {
-				/* tslint:disable-next-line:no-eval */
-				eval(evalString);
-			} catch (e) {
-				console.error("content eval failed: \neval string: %s\n error message: %s", evalString, e);
+		// const [table, offerId, column, jsonPath] = getCsvKeyElements(csvKey);
+		const [table, offerName, column, jsonPath] = getCsvKeyElements(csvKey);
+		const currentOffer = _.find(allOffers, function(offer: Offer) {
```
liorama: Why lodash and not just `allOffers.find()`?

If you are doing this just to get the offer id in-order to get to the offer content, you can just do one query and join the two tables together.

liorama: Also use arrow functions instead of function definistions.

scripts/src/admin/translations.ts
```diff
@@ -61,7 +68,8 @@ const EXCLUDED = [
 ];
 
 function constructRow(contentType: ContentType, key: string, str: string) {
-	const path = `${ contentType }:${ key.replace(KEY_TO_PATH_REGEX, "$1") }`;
+	const path = `${ contentType }:${ key.replace(/:(.*?):/, ":").replace(/\[.\]/g, "") }`;
```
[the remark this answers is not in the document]
shaharsol: added a comment instead, much more useful

scripts/src/admin/translations.ts
```diff
@@ -61,7 +68,8 @@ const EXCLUDED = [
...
+
```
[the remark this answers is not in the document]
shaharsol: done

scripts/src/admin/translations.ts
shaharsol: much more clear to use offerId as pkey

scripts/src/admin/translations.ts
shaharsol: done

scripts/src/admin/translations.ts
shaharsol: using js find instead. hard to get used to *not* using _

PR #371, opened 2019-08-13
remote repl with repl client

liorama:
Allow remote connection to a node REPL running on the environment's admin instance. Enjoy with caution.

### DIFF
--- a/scripts/src/repl-client.ts
+++ b/scripts/src/repl-client.ts
@@ -3,7 +3,7 @@ import * as net from "net";
 import * as path from "path";
 
 const args = process.argv.slice(2);
-if (args.length < 1 || !(args[0] as string).match(/^[\w\d.]{3,}:\d{1,5}$/)) {
+if (args.length < 1 || !(args[0] as string).match(/^[\w\d.\-_]{3,}:\d{1,5}$/)) {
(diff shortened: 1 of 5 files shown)


scripts/src/repl-client.ts
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env node
```
liorama: @victordoc @shaharsol (and @doodyparizada why not).
Funny anecdote, I had an internet outage today. Finally had time now to see what the problem was, entered my router's web interface, which showed that the internet connection is up and running. Yet nothing loads. I try to ping 8.8.8.8 (and 1.1.1.1) from my computer but hit an ICMP request timeout (so there's a connection but something is wrong). Since the the router says it's connected I wanted to try and ping one of the above IPs from it. The problem, MacOS does not come with telnet preinstalled and I don't have an internet connection to install it. I desperately need a telnet like TCP client, AND that's when I remembered this simple client I wrote a week or so ago. Which worked magnificently! exactly how I would expect the Telnet client to (I even installed and compared them, the only difference was a few characters this client prints out when the connection is initially made). So, in the future you can use this client with port 23, and it should do the trick. That's it, you don't have to install telnet now. Of course, you can also use Telnet instead of this client to connect to the remote REPL (just specify the correct port)

PR #373, opened 2019-08-18
Shaharsol/gradual migration kill switch

shaharsol:
The combination of the kill switch and gradual migration.
If you update the bc version from 2 to 3 in the admin, 3 more params are pushed to the app config (you need to refresh the admin page to see them): `gradual_migration_date`, `hourly_migration`, and `minute_migration`. `gradual_migration_date` is initialised to now, `hourly_migration` and `minute_migration` to 1 and admin is expected to manually increase them.

It is then the job of `/migration/info` and `user authentication` to check if migration is needed and to prevent it if limits have been reached.


scripts/src/admin/services.ts
```diff
@@ -562,6 +563,14 @@ export const updateAppConfig = async function(req: UpdateAppConfigRequest, res:
 	}
 
 	const app = (await Application.findOne({ id: req.params.application_id }))!;
+
+	// if the change is migrating from v2 to v3, enforce gradual migration
+	if (app.config.blockchain_version === "2" && config.blockchain_version === "3"){
+		config.gradual_migration_date = moment().toISOString();
+		config.limits.hourly_migration = 1;
```
Doc999tor: You don't really want to add these values to production - we should get it from the config

Doc999tor (changes requested):

scripts/src/admin/services.ts
shaharsol: done

Doc999tor (dismissed):

Doc999tor (approved):

